PR #1, opened 2019-01-28
Add custom footer, add links in default footer

RemingtonReackhof:

### DIFF
--- a/src/Layout/Footer/Footer.component.tsx
+++ b/src/Layout/Footer/Footer.component.tsx
@@ -107,6 +107,18 @@ interface FooterState {
   currentBreakpoint: string;
 }
 
+interface ItemChild {
+  key: number;
+  href: string;
+  label: string;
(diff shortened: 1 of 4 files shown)


sharkparty (approved):
I think this solves the immediate need nicely and allows for us to move forward. I have some tangential thoughts, but let's stay focused on the objectives. Can you take a peak at the changes needed for the header and begin on those @RemingtonReackhof ? This is good to merge whenever you're ready.

RemingtonReackhof:
@sharkparty I'll merge it now and will take a look at the header. Where can I get the rundown for what's needed for that?

sharkparty:
I actually don't know. Your team is driving the requirements for that, so AFAIK they want the ability to hide the search bar—which I think I did (`disableSearch={true}`)—and the ability to customize the links in both menus, which should be supported.

PR #2, opened 2019-01-28
feature(Input) Adding initial commit of input

Michelle-shw:
Adding initial commit for input

### DIFF
--- a/src/Form/Input/Input.component.tsx
+++ b/src/Form/Input/Input.component.tsx
@@ -3,19 +3,21 @@ import * as React from 'react';
 // VENDOR
 import classNames from 'classnames';
 import styled from 'styled-components';
+import { get } from 'lodash';
 // THEME
-import { fonts, colors } from '../../theme';
+import { fonts, colors, variables } from '../../theme';
 
 interface InputProps {
   className?: string;
+  inputType?: any;
   label?: string;
   placeholder?: string;
   disabled?: boolean;
   defaultValue?: any;
   value?: any;
+  onChange?: any;
   // onFocus?: () => void;
   // onBlur?: () => void;
-  onChange?: (e: any) => void;
   validate?: (e: any) => void;
   format?: (e: any) => void;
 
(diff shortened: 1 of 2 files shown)


src/Form/Input/Input.component.tsx
```diff
@@ -62,10 +57,28 @@ export const StyledInput = styled.div `
     }
   }
 
+  .field-input {
+    flex: 1;
+    overflow: hidden;
+    width: 100%;
+    padding: 24px 0 6px ${({ prefix }: InputProps) => !prefix ? '.75rem' : 'calc(1rem - 3px)'};
```
sharkparty: This doesn't seem quite right.

Syntax:
`[rule]: ${(props) => props ? 'value if props' : 'value if !props'};`

Here:
`padding: 24px 0 6px ${({ prefix }: InputProps) => !prefix ? '.75rem' : 'calc(1rem - 3px)'};`

We define a padding for everything but the `padding-left`. I think I'd prefer:

`padding: ${({ prefix }: InputProps) => !prefix ? '1.5rem 0 0.375rem .75rem' : '1.5rem 0 0.375rem calc(1rem - 3px)'};`

This solves mixed units and it's a little bit easier to read. It's not wrong, though. This just seems tidier. What do you think?

src/Form/Input/Input.component.tsx
```diff
@@ -24,36 +25,30 @@ interface InputProps {
...
     position: relative;
-    font-size: calc(1rem - 2px);
-    font-weight: 500;
-    background: none;
-    z-index: 1;
+    display: flex;
+    flex-flow: row wrap;
+    border-radius: ${variables.borderRadius};
+    border: solid 1px ${colors.neutrals.ash.light};
+    opacity: ${({disabled}: InputProps) => disabled && '0.5'};
```
sharkparty: This is something I need to change in the `tsconfig`, but can we be consistent with the space in destructuring of the props from an object. I prefer `({ disabled })` over `({disabled})`, but I'm open to feedback. Whatever we decide, I'll add to the tsconfig and prettier so we don't have to worry about this in the future.

src/Form/Input/Input.component.tsx
sharkparty: Should we define the width property using this calculation as well? Otherwise, it could be 100% + the padding values and overflow parent elements.

src/Form/Input/Input.component.tsx
```diff
@@ -62,10 +57,28 @@ export const StyledInput = styled.div `
     }
   }
 
+  .field-input {
+    flex: 1;
+    overflow: hidden;
+    width: 100%;
+    padding: 24px 0 6px ${({ prefix }: InputProps) => !prefix ? '.75rem' : 'calc(1rem - 3px)'};
+    font-size: calc(1rem - 2px);
```
sharkparty: Do we really want a 14px font-size here? `0.875rem` is easier if we do.

src/Form/Input/Input.component.tsx
```diff
@@ -74,11 +87,12 @@ export const StyledInput = styled.div `
     margin-top: calc(1rem - 1px);
     top: 0;
     font-size: calc(1rem - 2px);
-    font-weight: 500;
+    font-weight: ${baseFontWeight};
     line-height: 1.29;
```
sharkparty: 1.29 what?

src/Form/Input/Input.component.tsx
Michelle-shw: This width is only for the input part of this component. Currently 100% means it's 100% to the `container - prefix`. Why do you think it can be overflow here?

sharkparty: This component has a width of 100% and a padding applied which makes it larger than a containing element I think. [Reference 1](https://stackoverflow.com/questions/4879788/padding-adds-to-div-width-height)

knksmith57: what's the idea re: _not_ opting into `border-box` sizing?

sharkparty: Asking the real questions. I had originally added it to the normalize global override, but I wasn't sure what our internal teams would be familiar with. I'd gladly re-introduce this if it makes calculation easier for our eng team, but not if it creates another obstacle. @knksmith57 and @Michelle-shw what's your preference here? Consider this an informal poll. =P

knksmith57: unless existing code relies specifically on `content-box` behavior (which would surprise me), my gut says we should absolutely prefer `border-box`.

sharkparty:
Hey @Michelle-shw can you update this branch so we can merge it?

PR #15, opened 2019-03-21
Remove lodash

sharkparty:

- [x] Used [conventional commits](https://www.conventionalcoamits.org/) for all work.

---------

Updated all dependencies to be latest.
Removed lodash because it's not really needed for this project.


utils/get.ts
```diff
@@ -0,0 +1,2 @@
+export const get = (obj: any, path: string, defaultValue: any) => path.split('.')
+  .reduce((a, c) => (a && a[c] ? a[c] : (defaultValue || null)), obj);
```
zero0Halo: Since the function is written so briefly it's not really self documenting. It'd be good to have a jsdoc block above this to explain what it's for, since although we may know it's replacing lodash's `get` no one else would know.

sharkparty: 👍

sharkparty: Added a spec and docs.

utils/get/get.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @param obj - an object that needs to be safely traversed `{ a: { b: true } }`
+ * @param path - a string that represents the path of the object `'a.b'`
+ * @param defaultValue - a value to return if the path doesn't resolve in the object
+ */
+export const get = (obj: any, path: string, defaultValue?: any) => path.split('.')
```
sharkparty, line 7: @zero0Halo bc I unit tested this, I noticed that I omitted the default being an optional prop. thanks for the push!

zero0Halo: One last change. Put a a description as to what the function is for on line 2. I am a comment nazi. Plain english, ex: "Our version of lodash's get() method, just less bloated" or something to that effect.

sharkparty: Nice catch.

zero0Halo: Looks good!

zero0Halo (approved):
Looks good.

PR #29, opened 2019-04-03
Feature: Layout

zero0Halo:
feature(page/layout): updated the existing page component to serve as layout 

feature(info): enabled storybook info add on to help auto-generate documentation

---------

My code does...


sharkparty:
checks are failing on this. Can we resync this with master and get the CI to pass? I'll still review it, but I imagine it'll be changing as I review...

.storybook/config.js
```diff
@@ -1,10 +1,14 @@
 // STORYBOOK
 import { configure, addDecorator } from '@storybook/react';
 // ADDONS
+import { withInfo } from '@storybook/addon-info';
 import { withA11y } from '@storybook/addon-a11y';
 import { withKnobs } from '@storybook/addon-knobs';
 import { withViewport } from '@storybook/addon-viewport';
 
+addDecorator(withInfo({
+  source:false,
```
sharkparty, line 12: Super-excited about this!

src/Icon/Icon.component.spec.tsx
```diff
@@ -4,26 +4,20 @@ import * as renderer from 'react-test-renderer';
...
 // ENZYME
-import { shallow, mount } from 'enzyme';
+import { mount } from 'enzyme';
 
 // TEST SETUP
 const subject = <Icon type="hamburger" />;
-const component = shallow(subject);
 
-describe('Component: Button', () => {
+describe('Component: Icon', () => {
```
sharkparty, line 12: Thank you!

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    margin: 0 auto;
 `;
 
-const StyledBodyElement = styled.section`
-    flex: 1 0 auto;
-    background-color: ${colors.silver.base};
-    max-width: 100vw;
-    align-self: stretch;
-    padding: ${sizes.padding.lg} ${sizes.padding.xl};
+const StyledContent = styled.div<StyledContentProps>`
```
sharkparty: We may wanna use a `main` here in place of a `div` for SEO and best practices: https://www.w3schools.com/tags/tag_main.asp

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    max-width: ${props => props.contentWidth};
+    background-color: ${props => props.contentBackgroundColor};
+    margin: 0 auto;
 `;
 
+// Interfaces
+// ------------------------------------------------------------------------------------------------------------------
 interface PageProps {
+    /** The header of the website. This can be a component. */
+    header?: any;
```
sharkparty, line 29: I don't know the answer to this question, but is there a way we can enforce only the right element types AND use `children`? My only hack to finding out how to do this is to type-cast the props (which works with what you have here), but I'd like to be able to pass all these elements in as children and have them all work 'automatically'.

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    /** The width of the content within the layout. */
+    contentWidth?: string | number;
+    /** The width of the area for the sidebar. */
+    sidebarWidth?: string | number;
+    /** If the sidebar is on the left or the right of the layout */
+    sidebarAlign?: SidebarAlignment;
+    /** The actual sidebar. This can be a component. */
+    sidebar?: any;
+    /** The background color of the layout. Default is transparent. */
+    layoutBackgroundColor?: string;
```
sharkparty: This prop and the subsequent one can be a slippery slope. I'm kosher leaving them here, but the second I start hearing about needing font-color, anchor tag treatment, etc. overrides, I'll yank these props out of here and ask people to extend this using classes.

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    /** The background color of the content within the layout area. Default is transparent. */
+    contentBackgroundColor?: string;
+    /** Additional classname. */
+    className?: string;
+    children?: any;
+}
+
+interface SidebarLayoutProps {
+    /** The width of the layout, edge to edge. */
+    layoutWidth?: string | number;
```
sharkparty: Do are you policing the units here? If you allow only a number then everything can always be in REMs, but it's not quite as intuitive for downstream users. If you allow strings, then people can pass in anything and that can get unwieldy too. I'm not positing a preference, here, but more asking what your thoughts are on this?

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
 }
 
-export const Page = ({ children }: PageProps): React.ReactElement<any> => (
-    <StyledPageElement className="anchor-page">
+interface StyledLayoutProps {
+    layoutWidth?: string | number;
+    layoutBackgroundColor?: string;
+}
+
+interface StyledContentProps {
```
sharkparty: Can we break all of these components apart?

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    contentWidth?: string | number;
+    contentBackgroundColor?: string;
+}
+
+// Components
+// ------------------------------------------------------------------------------------------------------------------
+export const Page = (props: PageProps): React.ReactElement<any> => (
+    <StyledPage className={classnames('anchor-page', props.className)}>
         <NormalizeCSS />
         <GlobalCSS />
```
sharkparty: I'm debating on having `GlobalCSS`. What are your thoughts on this?

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
-        <Header />
-        <StyledBodyElement>{children}</StyledBodyElement>
-        <Footer />
-    </StyledPageElement>
+        <Grid
+            columns={1}
+            rows={`minmax(${props.headerHeight},auto) 1fr minmax(${
```
sharkparty: Did this pass the prettier/linter?

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+            {props.header && <Cell>{props.header}</Cell>}
+
+            <Cell>{props.children}</Cell>
+
+            {props.footer && <Cell>{props.footer}</Cell>}
+        </Grid>
+    </StyledPage>
+);
+
+Page.defaultProps = PageDefaultProps;
```
sharkparty: Let's talk about default props. I think we need a standard here and I wanna hear your input. I'd either like all defaults to be in the method signature, or all defaults like this. My opinion is that this isn't quite as readable for the average person, but I do wanna chat about it from other perspectives.

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+
+const DefaultLayout = (
+    props: DefaultLayoutProps
+): React.ReactElement<any> => (
+    <StyledLayout
+        layoutWidth={props.layoutWidth}
+        layoutBackgroundColor={props.layoutBackgroundColor}
+        className={classnames('anchor-default-layout', props.className)}
```
sharkparty: Does this need to be `default-layout` or can it just be layout? It's our only layout so default is kind of implied IMO.

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+    <StyledLayout
+        layoutWidth={props.layoutWidth}
+        layoutBackgroundColor={props.layoutBackgroundColor}
+        className={classnames('anchor-sidebar-layout', props.className)}
+    >
+        <StyledContent
+            contentWidth={props.contentWidth}
+            contentBackgroundColor={props.contentBackgroundColor}
+        >
+            {props.sidebarAlign === RIGHT ? (
```
sharkparty: Ahh, typescript shows a little value here: we can use strict props to make binary configuration out of strings without having to have messy conditional blocks that check both sides of the condition.

src/Layout/Page/Page.component.tsx
```diff
@@ -2,50 +2,215 @@
...
+}
+
+interface StyledContentProps {
+    contentWidth?: string | number;
+    contentBackgroundColor?: string;
+}
+
+// Components
+// ------------------------------------------------------------------------------------------------------------------
+export const Page = (props: PageProps): React.ReactElement<any> => (
```
sharkparty: Can we destructure all props?

src/Layout/Page/Page.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+  // DEFAULT_SIDEBAR_WIDTH,
+  // TRANSPARENT,
+} from './Page.component';
+
+describe('Component: Page', () => {
+  const subject = <PageComponent />;
+  const wrapper = mount(subject);
+  const component = shallow(subject);
+
+  it('should be defined.', () => {
```
sharkparty: `'should match it's default snapshot.'`

src/Layout/Page/Page.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+  DEFAULT_HEADER_HEIGHT,
+  DEFAULT_FOOTER_HEIGHT,
+  DEFAULT_LAYOUT_WIDTH,
+  TRANSPARENT,
+  DEFAULT_CONTENT_WIDTH,
+  // DEFAULT_CONTENT_WIDTH,
+  // DEFAULT_SIDEBAR_WIDTH,
+  // TRANSPARENT,
+} from './Page.component';
+
```
sharkparty, line 11: `const {  DefaultLayout  } = Page;` can make things easier.

src/Layout/Page/Page.stories.tsx
```diff
@@ -3,13 +3,80 @@ import * as React from 'react';
...
+    background: ${colors.white.base};
+`;
+
+// Stories
+// ------------------------------------------------------------------------------------------------------------------
+storiesOf('Components/Layout/Page', module).add('Default Layout', () => (
+    <StyledStory>
+        <PageComponent header={<Header />} footer={<Footer />}>
+            <Page.DefaultLayout
+                layoutWidth={text('Layout Width', DEFAULT_LAYOUT_WIDTH)}
```
sharkparty: This is so neat.

src/theme/GlobalStyles/Globals.styles.ts
```diff
@@ -22,6 +22,7 @@ export const GlobalCSS = createGlobalStyle`
     overflow-y: hidden;
     overflow-x: hidden;
     height: 100vh;
+    /* TODO: Change this back to #ffffff */
     background-color: #710a32;
```
sharkparty, line 26: Thanks!

sharkparty (changes requested):
Added a few comments to get the discussion going. HMU with questions or comments.

sharkparty:
Wow, github is really scary with its changes-requested treatment. That's intimidating and I don't like it.

src/Layout/Page/Page.component.tsx
zero0Halo: I think this is a good idea. I wasn't crazy about using `div` or `section` for the main wrapper. Ironically the `main` tag was looking me right in the face and I didn't even think of it.

src/Layout/Page/Page.component.tsx
zero0Halo: Using props is more standard/natural way of approaching the problem, and I think trying to not be too clever on this is the best way to go. Still, the idea of sniffing out a child's type and applying it correctly is an interesting notion. Definitely would like to talk about this because it's intriguing.

src/Layout/Page/Page.component.tsx
zero0Halo: Agreed. I put the layout/content background color values for common use cases I was seeing on RMN's site where the background colors were extending from edge to edge, but the content was in the center. Granted that could be taken care of solely in the user's own component, but it got me thinking that perhaps giving them additional options might be good too. 

But yeah, it does potentially open a door for an onslaught of options.

src/Layout/Page/Page.component.tsx
zero0Halo: I had been questioning having string/number as the type almost from the get-go. I think I'm going to yank `number` and keep `string` because ultimately that allows for more flexibility and less confusion.

src/Layout/Page/Page.component.tsx
zero0Halo: That's my plan. I'm going to do it in an upcoming push.

src/Layout/Page/Page.component.tsx
zero0Halo: I did most of the work on this w/o it, so I'm fine with ditching it. One possibility, a prop to enable it?

src/Layout/Page/Page.component.tsx
zero0Halo: Yep!

src/Layout/Page/Page.component.tsx
zero0Halo: I'm up for the discussion, I'm definitely not married to the way I have it.

src/Layout/Page/Page.component.tsx
zero0Halo: There's two layouts, DefaultLayout and SidebarLayout. I went this route because this naming gives us latitude if we wanted to add more layouts. I think the door should be left open for that possibility, because although it's all we have right now that doesn't mean the list won't get bigger.

src/Layout/Page/Page.component.tsx
zero0Halo: I'd like to talk about this more as I'm not sure how to go about that.

src/Layout/Page/Page.component.tsx
zero0Halo: Yeah that should be fine. Originally I was spreading `props` to the styled components, but I stopped doing that.

src/Layout/Page/Page.spec.tsx
zero0Halo: Since I'm going to breaking up the components this shouldn't be an issue (soon).

src/Layout/Page/Page.stories.tsx
zero0Halo: Gotta love those knobs.

src/Layout/Page/Page.component.tsx
sharkparty: Yeah, I often lose sight of the trees in the forrest. I'm pretty sure the content area for Rx is inside of a `section` 🙈

src/Layout/Page/Page.component.tsx
sharkparty: Maybe we add a v2 ticket. This can be outside of MVP, but still worth considering, right?

src/Layout/Page/Page.component.tsx
sharkparty: For some of those we can create a `Hero` component that functions like Bootstrap's Hero, but this also solves that nicely. I just wanna get ahead the 'onslaught', but I don't think that merits a change to this. It's a nice feature to have.

src/Layout/Page/Page.component.tsx
sharkparty: SGTM. I did have Icons implemented the other way, but IIRC Icons now take in a scale prop that's an Enum. Can you do me a favor and check if `Typography` is consistent with this pattern? I think it only takes in a `number` and it assumed that that number is in REMs. If so, can you create a bug ticket and assign it to me to make that consistent with this approach?

src/Layout/Page/Page.component.tsx
sharkparty: I don't think it'll define anything anymore TBH. I created a Typography component that renders all permutations of our type, so we no longer need to have typography in there and that's all that was there. The Normalize is probably the only thing we need and IDK if that will be essential as we build out components with known un-normalized CSS we can just build that into the component.

src/Layout/Page/Page.component.tsx
sharkparty: I'll add a mtg to discuss this. I'll also try and get a few others in the room to talk about it.

src/Layout/Layout.component.tsx
```diff
@@ -0,0 +1,80 @@
...
+    className?: string;
+    children?: any;
+}
+
+interface StyledLayoutProps {
+    layoutWidth?: string;
+    layoutBackgroundColor?: string;
+}
+
+interface StyledContentProps {
```
sharkparty, line 51: Do we want to extricate `Content` from `Layout` as a nested component in a nested dir?

src/Layout/Layout.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+        expect(testSubject.find('StyledContent').prop('contentWidth')).toEqual(
+            contentWidth
+        );
+        expect(
+            testSubject.find('StyledContent').prop('contentBackgroundColor')
+        ).toEqual(contentBackgroundColor);
+    });
+});
+
+// tslint:disable-next-line: no-console
```
sharkparty: This lint disabler and the console.log can be removed.

src/Layout/Layout.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+                )}
+            >
+                <h1>Default Layout</h1>
+                <h2>This example uses the Page component as a wrapper.</h2>
+            </Layout>
+        </Page>
+    </StyledStory>
+));
+
+// storiesOf('Components/Layout/Page', module)
```
sharkparty: If we don't want these stories, then let's delete them. Add the readme scaffold, though. That we'll likely want downstream.

src/Layout/Page/Page.stories.tsx
```diff
@@ -3,21 +3,50 @@ import * as React from 'react';
...
+import {
+    Page,
+    DEFAULT_HEADER_HEIGHT,
+    DEFAULT_FOOTER_HEIGHT,
+} from './Page.component';
+import { Footer } from '../Footer/Footer.component';
+import { Header } from '../Header/Header.component';
+import { colors } from '../../theme';
 // README
-import * as README from './README.md';
```
sharkparty: We'll prolly wanna keep the readme, but lose the other commented out code.

src/Layout/SidebarLayout/SidebarLayout.component.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+interface StyledContentProps {
+    contentWidth?: string;
+    contentBackgroundColor?: string;
+}
+
+// Components
+// ------------------------------------------------------------------------------------------------------------------
+export const SidebarLayout = ({
```
sharkparty, line 74: Finest possible hair being split here: Do we need the added `Layout` for this component name? It sounds redundant, but I will definitely not die on this hill.

zero0Halo: I think w/o the word 'Layout' the usage of the component could be confusing. Dropping `<SidebarLayout>` in jsx is much more clear than `Sidebar` IMO

src/Layout/Layout.stories.tsx
zero0Halo: That was me being dumb. Never meant to leave the commented garbage.

src/Layout/Layout.component.tsx
zero0Halo: The reason I have Layout and Content styled components is to more easily be able to change the width of both of them via the single Layout (or SidebarLayout) component(s). I don't think it's necessary (right now anyways) to make them individual components for the user to consume. Breaking them up like that removes some of the simplicity of just dropping in `Layout` or `SidebarLayout`. I'm open to discuss this more though, as perhaps I'm not seeing the same angle as you.

sharkparty: 👍

sharkparty (approved):
LGTM

PR #33, opened 2019-04-10
chore: add color documentation

bjohnsonrmn:

---------

My code adds a color swatch component to the doc site to show the colors in our theme file. I also made a few edits to the `contributing.md` and I moved a few styles that are specific to the doc homepage so that they apply to that page only.


src/theme/GlobalStyles/Globals.styles.ts
```diff
@@ -19,11 +19,6 @@ export const GlobalCSS = createGlobalStyle`
     font-size: ${fonts.baseFontSize}px;
     letter-spacing: 0;
     line-height: 1.357142857rem;
-    overflow-y: hidden;
-    overflow-x: hidden;
-    height: 100vh;
-    /* TODO: Change this back to #ffffff */
```
bjohnsonrmn, line 25: I moved this CSS to the index page because it seems to apply only there. Should I preserve this `TODO` comment, or does this change resolve it?

CONTRIBUTING.md
```diff
@@ -72,7 +72,7 @@ $ npm i
 ## Run the app
 
 ```sh
-$ yarn dev
+$ yarn storybook
```
brandonlilly, line 75: Nice catch 👍 There's also the `npm run dev` command below that'll need to get updated

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <input
+          type="radio"
+          id="copyChoiceHexCode"
+          name="copyChoice"
+          value="hexCode"
+          checked={copyChoice == 'hexCode'}
+          onClick={ () => setCopyChoice('hexCode')}
+        >
+        </input>
+        <label for="copyChoiceHexCode"> Hex Code</label>
```
brandonlilly: React is giving a warning that the label doesn't work. It should be `htmlFor=`. But I prefer wrapping the element with the label over using `for`/`id`s anyway:
```html
        <label>
          <input type="radio" name="copyChoice" value="hexCode"
          /> Hex Code
        </label>
```
```jsx
        <label>
          <input type="radio" name="copyChoice" value="variableName"
          /> Variable Name
        </label>
```

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
+// REACT
+import * as React from 'react';
+// VENDOR
+import styled, { css } from 'styled-components';
```
brandonlilly: `{ css }` doesn't appear to be used

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+import classNames from 'classnames';
+import { ColorContrastHelpers } from '../ColorContrastHelpers';
+import { ColorSwatch } from '../ColorSwatch';
+// THEME
+import { colors, sizes } from '../../../../src/theme';
```
brandonlilly: `colors` also doesn't appear to be used

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+
+    <CopyToClipboard text={copyToClipboardText}>
+      <StyledColorSwatch
+        className={classNames(className)}
+        backgroundColor={backgroundColor}
+        textColor={textColor}
+        size={size}
+      >
+       <p class="ColorSwatchName">{name}</p>
```
brandonlilly: `className`

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+  contrastOnF1F1F1,
+  contrastOnFAFAFA,
+  contrastOnWhite,
+  copyChoice,
+  name,
+  size,
+  textColor
+}: ColorSwatchProps): React.ReactElement<any> => {
+
+  let copyToClipboardText = backgroundColor;
```
brandonlilly: just a suggestion, feel free to ignore
```js
const copyToClipboardText = copyChoice == 'variableName' ? backgroundColor : name
```

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  ${({ size = 'large' }: ColorSwatchProps) => ColorSwatchSizeStyles[size]}
+`;
+
+
+// JSX MARKUP
+export const ColorSwatch = ({
+  className,
+  backgroundColor,
```
brandonlilly: In my opinion, `ColorSwatch` should take a `color` or something similarly named (e.g. `swatchColor`, `themeColor` etc) as a prop to go along with the swatch `name`, which is then used as the background color, instead of specifying a backgroundColor explicitly. 

That way the way in which the swatch decides to present the swatch color is decoupled from its interface.

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+    baseColorHexCode={baseColorHexCode}
+    className={classNames(className)}
+    darkColorHexCode={darkColorHexCode}
+    lightColorHexCode={lightColorHexCode}
+  >
+
+    <ColorSwatch
+      className={classNames(className)}
+      backgroundColor={baseColorHexCode}
+      contrastOnBlack={ColorContrastHelpers.getRoundedColorContrast(baseColorHexCode, '#000000')}
```
brandonlilly: I also think that the `contrastOnX` and `textColor` props should instead be calculated in the child `ColorSwatch`, rather than calculated here. All of them are just a function of the backgroundColor, so it wouldn't make sense to pass in any other values for a given color.

That way it's also much easier to change which colors we decide to contrast against in the future, and you aren't stuck doing this math every single time you want to use a `ColorSwatch`.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+  `,
+  small: css`
+    .ColorSwatchPillContainer {
+      padding-top: 50px
+    }
+  `,
+}
+
+const StyledColorSwatch = styled.div`
+  background: ${props => props.backgroundColor };
```
brandonlilly: We should add some indication that the ColorSwatches are clickable- ie. `cursor: pointer`.

Nice to haves:
 * A copy clipboard icon on hover either in the corner or next to the `name` or `color` that will be copied (depending on copyChoice).
 * Some small visual indication when text is copied

But don't let me feature creep you

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export class ColorContrastHelpers {
+
+  public static getRoundedColorContrast(colorHexCode1, colorHexCode2): number {
+    let exactContrast = wcagContrast.hex(colorHexCode1, colorHexCode2);
+    return Math.round(exactContrast * 100) / 100;
+  }
+
+  public static getSuccessOrErrorColor(colorContrast): number {
+    if (colorContrast > 4.5) {
```
sharkparty: In cases where you're using a single predicate, you can use a ternary to abbreviate this a bit.
```
return 1 > 2 ? true : false;
```

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  public static getHigherContrastColor(colorHexCode): string {
+    let contrastOnBlack = wcagContrast.hex('#000', colorHexCode);
+    let contrastOnWhite = wcagContrast.hex('#fff', colorHexCode);
+
+    if (contrastOnWhite > contrastOnBlack) {
+      return '#fff';
+    } else {
+      return '#000';
+    }
```
sharkparty: All of this can be:
```
return wcagContrast.hex('#fff', colorHexCode) > wcagContrast.hex('#000', colorHexCode) ? '#fff' : '#000';
```

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+  color: ${props => props.textColor };
+  padding: 24px 16px 16px 16px;
```
sharkparty: We can use `REM`s here (1rem = 16px) and we can abbreviate this CSS: `padding 1.5rem 1rem 1rem;`

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+interface ColorSwatchStylesGroup {
+  [key: string]: FlattenSimpleInterpolation;
+}
+
+
+// STYLES
+const ColorSwatchSizeStyles: ColorSwatchStylesGroup = {
+  large: css`
+    .ColorSwatchPillContainer {
```
sharkparty: All css classes should be kabob-case. `color-swatch-pill-container`

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
sharkparty: +1 on @brandonlilly's comment, but also: I built out a new component called Typography that can help do all the text formatting automatically. I'm going to dump this here with no explanation to see how effective Stories and IDE autocomplete is at telling a developer how to use it. If you hit turbulence using this, then blow me up immediately and that's what will inform what we document.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+      padding-top: 200px
```
sharkparty: Why not use `height` to define a flex-box (FB) block element and then flex-end the content? Also, FB is new so if you haven't played with it, it's super cool.

sharkparty: Also, always use `REM`'s plz.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <CopyToClipboard text={copyToClipboardText}>
+      <StyledColorSwatch
+        className={classNames(className)}
+        backgroundColor={backgroundColor}
+        textColor={textColor}
+        size={size}
+      >
+       <p class="ColorSwatchName">{name}</p>
+       <p class="ColorSwatchBackgroundColor">{backgroundColor}</p>
+       <div class="ColorSwatchPillContainer">
```
sharkparty: You may be able to use `Grid` and `Cell`s to do this more easily.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+         <Pill value={contrastOnWhite} backgroundColor="#ffffff" textColor={backgroundColor}></Pill>
+         <Pill value={contrastOnFAFAFA} backgroundColor="#fafafa" textColor={backgroundColor}></Pill>
+         <Pill value={contrastOnF1F1F1} backgroundColor="#f1f1f1" textColor={backgroundColor}></Pill>
+         <Pill value={contrastOnBlack} backgroundColor="#000000" textColor={backgroundColor}></Pill>
```
sharkparty:
```
[
    { key: 1, value: contrastOnWhite, backgroundColor: '#fff', textColor: backgroundColor },
    ...
].map((pillData: PillDataItem) =>(<Pill {...pillData}>));
```

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
sharkparty: +1

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+  className?: string;
+  copyChoice: string;
+  darkColorHexCode?: string;
+  lightColorHexCode?: string;
+}
+
+
+// STYLES
+const StyledColorSwatchFamily = styled.div`
+  padding: 12px;
```
sharkparty: Same note on `REM`s

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+  margin: 24px;
+  width: 25%;
+  min-width: 300px;
+  display: inline-block;
```
sharkparty: Another shameless plug for flex-box.

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+      contrastOnF1F1F1={ColorContrastHelpers.getRoundedColorContrast(baseColorHexCode, '#F1F1F1')}
+      contrastOnFAFAFA={ColorContrastHelpers.getRoundedColorContrast(baseColorHexCode, '#FAFAFA')}
+      contrastOnWhite={ColorContrastHelpers.getRoundedColorContrast(baseColorHexCode, '#FFFFFF')}
+      copyChoice={copyChoice}
+      name={baseColorName+'.base'}
+      textColor={ColorContrastHelpers.getHigherContrastColor(baseColorHexCode)}
+    ></ColorSwatch>
+
+    { lightColorHexCode && (
+      <ColorSwatch
```
sharkparty, line 48: I think we can clean this up a little bit by separating the color swatch JSX from the logic determining what to render. [Google's implementation](https://github.com/mui-org/material-ui/blob/next/docs/src/pages/style/color/Color.js) for this is way different, but even at a high level look at how little JSX-foo is going on here.

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+// INTERFACES
+interface ColorSwatchFamilyListProps {
+  className?: string;
+}
+
+
+// STYLES
+const StyledColorSwatchFamilyList = styled.div`
+
```
sharkparty: Doesn't look like we need this styled component.

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+}: ColorSwatchFamilyListProps): React.ReactElement<any> => {
+
+  const [copyChoice, setCopyChoice] = useState('hexCode');
+
+  return (
+    <StyledColorSwatchFamilyList
+      className={classNames(className)}
+    >
+
+      <p> Each color swatch has 4 pills representing the color contrast when used on these colors:</p>
```
sharkparty: If you import Typography this page will be a lot easier to deliver, unless I did a poor job building that component, in which case let's make it better!

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
sharkparty: The wrapping syntax here I can go either way on. Just an FYI in a production app you'd still need the `htmlFor` as well as `aria` tags to promote a11y, but this is the docs site, so I'm not going to police that here. Further, I'm working on the `Input` for Anchor right now so eventually you'll get some of this for free.

docs/src/components/Pill/Pill.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+  backgroundColor: string;
+  textColor: string;
+  value: number;
+}
+
+
+// STYLES
+export const StyledPill = styled.div`
+  color: ${props => props.textColor };
+  font-size: 16px;
```
sharkparty: Generic `REM`s comment

docs/src/pages/index.tsx
```diff
@@ -1,12 +1,22 @@
...
 // VENDOR
-import styled, { css } from 'styled-components';
+import styled, { createGlobalStyle, css } from 'styled-components';
 import { withPrefix } from 'gatsby';
 // COMPONENTS
 import { Page } from '../components/Layout';
 // THEME
 
+const IndexPageGlobalCSS = createGlobalStyle`
+  html,
```
sharkparty: This is a little alarming. What's this thing for?

sharkparty:
@bjohnsonrmn—great start here! This is really coming along nicely!

bjohnsonrmn:
@brandonlilly  and @sharkparty, thanks for all the good feedback. I'm planning to work through it today!

CONTRIBUTING.md
bjohnsonrmn: ✅ I'll update the `npm run dev`

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
bjohnsonrmn: I'm going to look into the `htmlFor` and `aria` because I'd like to practice doing it the right way.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+        className={classNames(className)}
+        backgroundColor={backgroundColor}
+        textColor={textColor}
+        size={size}
+      >
+       <p class="ColorSwatchName">{name}</p>
+       <p class="ColorSwatchBackgroundColor">{backgroundColor}</p>
+       <div class="ColorSwatchPillContainer">
+         <Pill value={contrastOnWhite} backgroundColor="#ffffff" textColor={backgroundColor}></Pill>
+         <Pill value={contrastOnFAFAFA} backgroundColor="#fafafa" textColor={backgroundColor}></Pill>
```
brandonlilly: Minor suggestion: I suspect the pills would look more pleasing if they were a uniform width. Currently the width is being determined by however many digits the value happens to be.

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
bjohnsonrmn: Found this: https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-radio-buttons

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
bjohnsonrmn: ✅ Will remove

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
bjohnsonrmn: I am using `colors` it on line 30:
`  box-shadow: 0px 0px 12px 4px ${colors.ash.light};`

But `sizes` looks like it isn't used 😂 so I will remove that.

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
bjohnsonrmn: ✅ Updated

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: @sharkparty the `Typography` component looks cool. Looking at the stories and the interface it seems intuitive to me how to use. I am running into 2 difficulties with it though.

1. I am setting the text color via a prop in the form of a hex code because it is dynamic based on color contrast. I have this prop available to pass in to `Typography` but it looks like its interface does not accept hex code values (also this seems like an edge case to me).

2. I'm wanting to zero out margin (like a normalize). Was that an intentional choice to leave that out?

I don't think either of these are worth making changes to `Typography` at the moment, but perhaps they are reasons for using classes for now, and I will use `Typography` for other simpler elements. What do you think?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: I like it ✅

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: @sharkparty yeah I've only played with it a little, which is why I used a different way. I'd like to learn it, but I don't think I'd want to hold this ticket up for that unless you think it is important.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: @sharkparty hmmm, I actually find that more difficult to read. Do you mind if I leave it?

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
bjohnsonrmn: ✅ I'll remove.

docs/src/pages/index.tsx
bjohnsonrmn: @sharkparty I moved those styles from `src/theme/GlobalStyles/Globals.styles.ts` to here because it seemed that they were necessary only on the index page (they seem to be for the wave animation).

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
bjohnsonrmn: I'm about to push up some code with `ColorSwatch` refactored to have less props, so that should clean up this file a bit. Let me know if this still seems like an issue to you. Also, when I first started this ticket, some of the colors only had `base` defined, which is why I did that check. Perhaps we could remove if that colors file has a public contract to have `light` and `dark` also defined.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: I like it. I updated to a uniform width ✅

bjohnsonrmn:
@sharkparty and @brandonlilly I just pushed fixes for most if not all the feedback, as noted in my comments, so it's ready for you to take another look at on Monday 👍

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
bjohnsonrmn: I initially forgot to remove this, but I just removed it ✅

bjohnsonrmn:
@sharkparty and @brandonlilly, here are a couple things to note on the work I did for Galen's redesign:

- Made variable name, color hex code, and rgba code individually clickable, which removed the need for the toggle between variable name and hex code
- Renamed ColorContrastHelpers to ColorHelpers because I added a function that wasn't related to contrast
- Re-worked the HTML/CSS structure of the color swatch component even though the props remained largely the same

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
```diff
@@ -0,0 +1,32 @@
+// VENDOR
+import * as wcagContrast from 'wcag-contrast';
+// THEME
+import { colors } from '../../../../src/theme';
+
+
+export class ColorContrastHelpers {
+
+  public static getRoundedColorContrast(colorHexCode1, colorHexCode2): number {
+    let exactContrast = wcagContrast.hex(colorHexCode1, colorHexCode2);
```
zero0Halo: Is there a reason for **let** instead of **const** here?

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
zero0Halo: I don't mind the verbosity. I got what was happening right away whereas the more terse line of code had me question my career choice.

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+  public static getSuccessOrErrorColor(colorContrast): number {
+    if (colorContrast > 4.5) {
+      return colors.success.base;
+    } else {
+      return colors.error.base;
+    }
+  }
+
+  public static getHigherContrastColor(colorHexCode): string {
+    let contrastOnBlack = wcagContrast.hex('#000', colorHexCode);
```
zero0Halo: I've seen this in a few spots now, so I will only comment one more time on it. Consider using **const** instead of **let**. Also since you are reusing string values (i.e. '#000') it would probably be better to assign those to a const to avoid magic numbers.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  if (copyChoice == 'variableName') {
+    copyToClipboardText = name;
+  }
+
+  return (
+
+    <CopyToClipboard text={copyToClipboardText}>
+      <StyledColorSwatch
+        className={classNames(className)}
```
zero0Halo: I might be missing it with the other comments breaking up the code, but is the usage of **classNames()** here necessary? There's no concatenation of classes that I can see, just the custom one the user can pass via props.

docs/src/components/ColorSwatchFamilyList/ColorSwatchFamilyList.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <label for="copyChoiceHexCode"> Hex Code</label>
+      </p>
+
+      <p>
+        <input
+          type="radio"
+          id="copyChoiceVariableName"
+          name="copyChoice"
+          value="variableName"
+          checked={copyChoice == 'variableName'}
```
zero0Halo: I've noticed it in a few spots but I'll mention it here. Generally best to use === instead of == for comparisons. This checks for type as well.

zero0Halo (commented):
I have some design questions & thoughts, but codewise I think Kam and Brandon hit a lot before I got here. I did make a few notes about const vs. let and === vs. == usage and also magic number usage.

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
bjohnsonrmn: No. I re-wrote this to use no variable at all on Kameron's suggestion (just a return).

docs/src/components/ColorContrastHelpers/ColorContrastHelpers.component.tsx
bjohnsonrmn: I think I was using `let` because I was changing the value later. I refactored this one as well to not have any variables. I did update to use named colors rather than hex codes.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: I think I'm just using 1 class for the styled components, so I think you are right that I may be able to remove `classNames()`. Let me check!

bjohnsonrmn: @zero0Halo this was a good catch. Not only was I able to remove `classNames()`, I removed the `className` prop as well! I think I thought that it was necessary for styled components somehow, but it seems it isn't according to these docs: https://www.styled-components.com/docs/basics#how-do-styled-components-work-within-a-component

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  .color-swatch-text-container p:hover {
+    background: ${colors.ash.light};
+    border-radius: 1rem;
+    cursor: pointer;
+    text-decoration: underline;
+  }
+
+  .temporary-copy-confirmation {
+    font-size: 1rem;
```
sharkparty, line 35: All of this can come from the Typography component that probably didn't exist when you started this work. @bjohnsonrmn I know you've got other things on your plate and this should have been called out earlier, so would you like me to refactor this with Typography, or would you like to work on that?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+// JSX MARKUP
+export const ColorSwatch = ({
+  colorHexCode,
+  colorName
+}: ColorSwatchProps): React.ReactElement<any> => {
+
+  const highContrastBorderColor = ColorHelpers.getHigherContrastColor(colorHexCode);
+  const colorRgbCode = ColorHelpers.getRgbColor(colorHexCode)
+
+  const [isCopied, setIsCopied] = useState(false);
```
sharkparty: Did you know you can type your hooks? Me neither until I did some digging:

```
useState<boolean>(false);
```

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [copyToClipboardText, setCopyToClipboardText] = useState(colorHexCode);
+  const handleCopy = function(copiedText) {
```
sharkparty: Do you need `function`-scope? or can this be an arrow function?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+    setIsCopied(true);
+    setTimeout(() => {setIsCopied(false)}, 1000);
+  }
+
+  return (
+    <StyledColorSwatch
+      colorHexCode={colorHexCode}
+      highContrastBorderColor={highContrastBorderColor}
+    >
+     {isCopied ? <span className="temporary-copy-confirmation">🎉 You copied "{copyToClipboardText}"</span> : null}
```
sharkparty: First emoji in the codebase!

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+     <CopyToClipboard text={colorHexCode} onCopy={() => handleCopy(colorHexCode)}>
+       <div className="color-swatch-pill-container">
+         <Pill value={ColorHelpers.getRoundedColorContrast(colorHexCode, '#ffffff')} backgroundColor="#ffffff" textColor={colorHexCode}></Pill>
```
sharkparty: Could this be:

```
{['#ffffff', '#fafafa', ...].map(() => (
    <Pill
        key={color}
        value={ColorHelpers.getRoundedColorContrast(colorHexCode, `${color}`)}
        backgroundColor={color}
        textColor={colorHexCode}
    />
))}
```

docs/src/components/Pill/Pill.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface PillProps {
+  backgroundColor: string;
+  textColor: string;
+  value: number;
+}
+
+
+// STYLES
+export const StyledPill = styled.div`
+  color: ${props => props.textColor };
```
sharkparty, line 20: We'll wanna set the `box-sizing` of this to `border-box` ([sauce](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)) to make sure all your padding and margins are predictably rendered without exploding a DOM block.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: I didn't, but that's cool. I'll update!

docs/src/pages/index.tsx
sharkparty: We should try to make them only impact this page and remove the global override, or leave them in there until we remove the splash page. There are definitely things in the globals just for the splash page, like that bright red BG, but I don't want to globalize that better, I want that entry to be refactored when we land the new index page.

src/theme/GlobalStyles/Globals.styles.ts
sharkparty: Commented above. We should chat about a good strategy for the splash page. I'm definitely convinced my temp solution wasn't ideal, but I still don't know what the best temp solution is.

sharkparty (changes requested):
I made requests for changes BUT since I was so late to the game, please tell me how much of this I can help own. Just LMK

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: @sharkparty I used `Typography` for some other elements, but this one seemed complex enough to deserve a class. I can re-work it, but are you saying every CSS attribute can be worked into `Typography` or only some of them?

sharkparty: The only ones that can't would be position, top, right, border, & bg. A few of those will be changing when we build out a composable base component. I can definitely be convinced this is better. Are those the only props and usage that's outside of Typo...y?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: I think this could be an arrow function 👍

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: We talked this over in person and decided to leave as is 👍

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: Yes! Will do.

docs/src/components/Pill/Pill.component.tsx
bjohnsonrmn: Thanks for this catch 🐟

docs/src/components/ColorHelpers/ColorHelpers.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  public static getHigherContrastColor(colorHexCode): string {
+    return wcagContrast.hex(colors.silver.base, colorHexCode) > wcagContrast.hex(colors.ash.light, colorHexCode) ? colors.silver.base : colors.ash.light;
+  }
+
+  public static getRgbColor(colorHexCode): string {
+    let h=colorHexCode.replace('#', '');
+    h =  h.match(new RegExp('(.{'+h.length/3+'})', 'g'));
+
+    for(var i=0; i<h.length; i++)
```
sharkparty: `for(let i=0; i<h.length; i++)`

docs/src/components/ColorHelpers/ColorHelpers.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+        h[i] = parseInt(h[i].length===1? h[i]+h[i]:h[i], 16);
+
+    return 'rgb('+h.join(',')+')';
```
sharkparty:
```
return `rgb(${h.join(',')})`;
```

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+  margin: 0 2rem 2rem 0;
+  width: 17rem;
+
+  .color-swatch-pill-container {
+    background: ${props => props.colorHexCode };
+    cursor: pointer;
+    padding: 7rem 1rem 1rem 1.5rem;
+  }
+
+  .color-swatch-text-container {
```
sharkparty: color-swatch-text-container can be its own styled component to house all these styles more neatly.

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    border-top: 0.05rem solid ${props => props.highContrastBorderColor};
+    padding: 1rem;
+  }
+
+  .color-swatch-text-container p {
```
sharkparty: This nested rule can be inside the curly brackets of the container:
```
color-swatch-text-container p {}
```
```
color-swatch-text-container {
      p { }
}
```

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    margin: 0;
+    padding: 0.25rem 0 0.25rem 0.5rem;
+  }
+
+  .color-swatch-text-container p:hover {
```
sharkparty: This can be nested very much the same way:

```
.color-swatch-text-container p:hover {}
```
can be
```
.color-swatch-text-container {
    p {
        &:hover {}
    }
}
```

sharkparty: Just remember:
> Note that objects should have at most 4 levels. Most of the time you'll stay around the 2-3 level range, and the fourth would be the interaction state.
~[Sauce](http://thesassway.com/beginner/the-inception-rule)

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    background: ${colors.ash.light};
+    border-radius: 1rem;
+    cursor: pointer;
+    text-decoration: underline;
```
sharkparty:
```
...
cursor: pointer;
text-decoration: underline;
...
```

If you're writing all of these styles that are defaults for the anchor tag, wouldn't it be better to use the anchor tag?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+              key={pillBackgroundColor}
+              value={ColorHelpers.getRoundedColorContrast(colorHexCode, `${pillBackgroundColor}`)}
+              backgroundColor={pillBackgroundColor}
+              textColor={colorHexCode}
+            />
+          ))}
+        </div>
+     </CopyToClipboard>
+     <div className="color-swatch-text-container">
+       <CopyToClipboard text={colorName} onCopy={() => handleCopy(colorName)}><Typography as="p" weight="700">{colorName}</Typography></CopyToClipboard>
```
sharkparty: This is now
```
<Typography as="p" weight="700">
```
this
```
<Typography tag="p" weight="700">
```
because of a bug identified by @brandonlilly

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+  baseColorHexCode: string;
+  baseColorName: string;
+  darkColorHexCode?: string;
+  lightColorHexCode?: string;
+}
+
+
+// STYLES
+const StyledColorSwatchFamily = styled.div`
+  margin: 1rem 0;
```
sharkparty: `box-sizing: border-box;`

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+    baseColorHexCode={baseColorHexCode}
+    darkColorHexCode={darkColorHexCode}
+    lightColorHexCode={lightColorHexCode}
+  >
+
+    <Typography as="h2" weight="600">{baseColorName}</Typography>
+
+    <ColorSwatch
+      colorHexCode={baseColorHexCode}
+      colorName={baseColorName+'.base'}
```
sharkparty:
```
colorName={`${baseColorName}.base`}
```

sharkparty: [sauce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+    ></ColorSwatch>
+
+    { lightColorHexCode && (
+      <ColorSwatch
+        colorHexCode={lightColorHexCode}
+        colorName={baseColorName+'.light'}
```
sharkparty: String template literal

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+      ></ColorSwatch>
+    )}
+
+    { darkColorHexCode && (
+      <ColorSwatch
+        colorHexCode={darkColorHexCode}
+        colorName={baseColorName+'.dark'}
```
sharkparty: String template literal

docs/src/pages/index.tsx
```diff
@@ -7,7 +7,11 @@ import { withPrefix } from 'gatsby';
 import { Page } from '../components/Layout';
 // THEME
 
+
 const StyledIndexPage = styled.div`
+    overflow-y: hidden;
+    overflow-x: hidden;
+    height: 100vh;
```
sharkparty, line 14: 👍

sharkparty (changes requested):
Minor tweaks and then heads up regarding new Typography API. May wanna rebase this branch.

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
bjohnsonrmn: The sauce was a fascinating read / sumptuous flavor journey 🍝

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
bjohnsonrmn: How are you determining which components need this?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: It looks like `Typography` was also updated to zero out margin and padding, which I was taking advantage of before. What is the preferred way to add in margin or padding?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: Will do. I've tasted this sauce before. It takes me back to Tuscany. 🍷

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: This is a good question. In this case, I would rather add a few anchor-tag-like qualities to `<p>` tags than use `<a>` and have to subtract a few anchor-tag-like qualities I don't want (blue color, inline display, etc.).

bjohnsonrmn:
@sharkparty I rebased and addressed all but two of your comments, on which I left questions for you.

docs/src/components/ColorSwatchFamily/ColorSwatchFamily.component.tsx
sharkparty: They all do, but containing components or anything that has padding/margins added usually catch my eye.

bjohnsonrmn: @sharkparty sounds good. Where's the sauce?

docs/src/components/ColorSwatch/ColorSwatch.component.tsx
bjohnsonrmn: ✅ I added this margin back in using the styled component.

sharkparty (approved):
LGTM

PR #38, opened 2019-04-19
Feature: Hero

brandonlilly:

---------

Adds a Hero component


src/Hero/__snapshots__/Hero.spec.tsx.snap
```diff
@@ -0,0 +1,23 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Component: Hero should be defined 1`] = `
+<section
+  className="anchor-hero sc-brqgnP dYemBD"
+  color="#FFFFFF"
```
brandonlilly: Using `color` as a prop is automatically adding it as an attribute. We could could keep the prop name as `color` and change what the color prop to the underlying styled component is called to fix this.

src/Typography/Typography.component.tsx
```diff
@@ -260,11 +268,13 @@ const StyledTypography = (as: Elements) => styled[as]`
 export const Typography = ({
     className,
     children,
-    as = 'span',
+    tag = 'span',
```
brandonlilly: As previously discussed, turns out `as` is part of the styled-components api, and using it when extending styles causes the "parent" styles to be dropped.

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,176 @@
...
+    text-align: ${props => props.align};
+    min-height: ${props => props.minHeight};
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+`;
+
+interface HeroProps {
+    background?: string;
+    align?: string;
```
brandonlilly: Change to `textAlign`?

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,176 @@
...
+        align={align}
+        {...props}
+    >
+        {children}
+    </StyledSubtitle>
+);
+
+Hero.Subtitle = HeroSubtitle;
+
+// Hero.Body
```
brandonlilly: Could've sworn I deleted this...

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,176 @@
...
+        weight={weight}
+        align={align}
+        size={size}
+        {...props}
+    >
+        {children}
+    </StyledTitle>
+);
+
+Hero.Title = HeroTitle;
```
brandonlilly, line 95: This sub components are importable both like:
```js
import { HeroTitle } from './Hero';
```
and:
```js
import { Hero } from './Hero';
const { Title } = Hero;
```
Anyone have opinions on if we should enforce just one way or the other?

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,176 @@
...
+        // margin: 0.375rem 0;
+    }
+`;
+
+export const HeroSubtitle = ({
+    className,
+    children,
+    scale = 16,
+    weight = 600,
+    color = 'inherit',
```
brandonlilly: I can remove this now that I've defaulted `Typography` to `inherit` anyway.

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,176 @@
...
+`;
+
+export const HeroTitle = ({
+    className,
+    children,
+    tag = 'h1',
+    weight = 600,
+    color = 'inherit',
+    align = 'inherit',
+    size = 2,
```
brandonlilly: This component should be an `h1` in almost all circumstances, but designs also show it as `2rem`. `Typography` warns that setting the size is a bad idea...

Maybe I should use both `tag` and `scale`

src/Hero/temporary/some_image.svg.tsx
```diff
@@ -0,0 +1,70 @@
+// REACT
+import * as React from 'react';
+
+/* tslint:disable max-line-length */
+export default () => (
+    <svg
```
brandonlilly, line 6: We need some dummy images that aren't IP

src/Typography/Typography.component.tsx
```diff
@@ -214,20 +214,28 @@ const ScaleTreatments = {
...
-    ${DefaultCSS[as]};
+    ${DefaultCSS[element]};
+
     // CSS Overrides
     text-align: ${({ align = 'inherit' }: any) => align};
     display: ${({ display }: any) => (display ? display : null)};
     text-transform: ${({ transform = 'none' }: any) => transform};
-    color: ${({ color = 'charcoal', hue = 'base' }: any) => colors[color][hue]};
+    color: ${({ color = 'charcoal', hue = 'base' }: any) =>
+        color === 'inherit' ? color : colors[color][hue]};
```
brandonlilly: This should be handled better in the future. I should add a comment

src/Typography/Typography.component.tsx
```diff
@@ -91,10 +91,10 @@ type DisplayValues =
 
 export interface TypographyProps {
     className?: string;
-    as?: Elements;
+    tag?: Elements;
     children?: any;
     weight?: FontWeights;
-    color?: Colors;
+    color?: 'inherit' | Colors;
```
brandonlilly: After discussion with @sharkparty, we thought it best to default Typography to inherit

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+interface StyledHeroProps {
+    align?: TextAlign;
+    background?: string;
+    color?: string;
+    minHeight?: string;
+}
+
+const StyledHero = styled.section<StyledHeroProps>`
+    background: ${props => props.background};
```
sharkparty, line 26: Prolly wanna add `box-sizing: border-box;` to this

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+    </StyledHero>
+);
+
+export default Hero;
+
+// Hero.Title
+// ------------------------------------------------------------------------------------------------------------------
+
+const StyledTitle = styled(Typography)`
+    text-shadow: 0 2px 4px rgba(0, 0, 0, 0.09);
```
sharkparty: These can be simplified to REM units. `0 0.125rem 0.25rem ...`

src/Hero/Hero.component.tsx
sharkparty: This is one of those practical over principle things IMO. For now, we'll allow tag overrides, but we should use this as a mechanism to enforce consistency with our friends on the design team.

src/Hero/Hero.component.tsx
sharkparty: Both is cool for these types of exports, but I have wholly deprecated default exports and will not allow them back. They break our bundler's chunking engine.

src/Hero/Hero.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+// Hero.Subtitle
+// ------------------------------------------------------------------------------------------------------------------
+
+const StyledSubtitle = styled(Typography)`
+    text-shadow: 0 2px 6px rgba(0, 0, 0, 0.04);
+    position: relative;
+    display: block;
+    && {
+        // margin: 0.375rem 0;
```
sharkparty: && block can be removed here

src/Hero/Hero.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Hero>
+        <Title>Hero Title</Title>
+        <Subtitle>Hero Subtitle</Subtitle>
+    </Hero>
+);
+const wrapper = mount(subject);
+const component = shallow(subject);
+
+describe('Component: Hero', () => {
+    it('should be defined', () => {
```
sharkparty, line 13: We should render a snapshot for (1) a hero with a header, (2) a hero with a header and a sub-header, & 3 a hero with custom content.

src/Hero/Hero.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+        /* tslint:disable */
+        console.log(subject);
+        console.log(wrapper.debug());
+        console.log(component.debug());
+
+        /* tslint:enable */
+        expect(subject).toBeDefined();
```
sharkparty, line 18: All these `toBeDefined` and `console` statements can be deleted once we have good tests in here. I use that console noise to gauge how well out testing is progressing—FYI.

src/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+import * as React from 'react';
+// STORYBOOK
+import { storiesOf } from '@storybook/react';
+import styled from 'styled-components';
+import { text, color } from '@storybook/addon-knobs';
+// ANCHOR
+import { Typography } from '..';
+import { colors } from '../theme';
+import * as README from './README.md';
+import SomeImage from './temporary/some_image.svg';
```
sharkparty, line 11: Let's create this inside the source code for this story. If you need help, LMK.

src/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+        return (
+            <StyledStory>
+                <Hero
+                    background={color('Hero Background', tealGradient)}
+                    color={color('Hero Color', colors.white.base)}
+                >
+                    <BreadCrumbs scale={14} weight={500}>
+                        {text('Subtext', 'Westeros / House Tyrell')}
+                    </BreadCrumbs>
+                    <Title>{text('Title', 'Olenna Tyrell')}</Title>
```
sharkparty, line 63: shaaaaaaame

src/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+// Styled Components
+// ------------------------------------------------------------------------------------------------------------------
+const StyledStory = styled.div`
+    background: ${colors.white.base};
+    width: 100vw;
+`;
+
+// Stories
+// ------------------------------------------------------------------------------------------------------------------
+const tealGradient = 'linear-gradient(224deg,#00cccd,#2de2b3)';
```
sharkparty: Few questions: 
- Are these standard colors?
- Do they exist in our `colors`?
    - What's the case for adding them?

src/Hero/Hero.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+            color: white;
+            border: thin solid white;
+            width: 9.5rem;
+            text-align: center;
+            margin: 0.1835rem auto;
+            font-size: 15px;
+        `;
+
+        const StyledImageWrapper = styled.div`
+            position: absolute;
```
sharkparty, line 80: This could use flex instead.

src/Hero/__snapshots__/Hero.spec.tsx.snap
sharkparty: You can prevent this by not spreading in all props, which is probably a good standard to adopt or by pulling out that prop by name to prevent it from being 'rested/spread' to the child component.

src/Hero/temporary/some_image.svg.tsx
sharkparty: This could come from Galen or Jen, depending on their schedule. I agree

src/Hero/Hero.stories.tsx
brandonlilly: I ripped this from the rmn banner, and didn't worry too much about having them being standard because this is just a backdrop for some stories.

It appears `#00cccd` is `tealBreaker.base`, and I don't think `#2de2b3` is standard.

src/Hero/Hero.stories.tsx
brandonlilly: How do you mean? display:flex in the parent component? I want the image to display beneath the title if it has to.

src/Hero/Hero.stories.tsx
brandonlilly: `dealEnvy.light` is pretty similar and in the new mockups I see, so I'll just switch it to that.

src/Hero/README.md
```diff
@@ -0,0 +1 @@
+# Hero
```
zero0Halo, line 1: Perhaps provide a usage example and explanation of props?

src/Hero/Hero.stories.tsx
brandonlilly: You mean defining the svg component inside the story that uses it? Seems like that'd add a lot of noise to this file.

src/Hero/README.md
brandonlilly: Added some barebones readme stuff

sharkparty (approved):
LGTM

PR #40, opened 2019-04-23
fix(Icon): Updated README, default color is inherited

zero0Halo:
chore: updated project w/yarn

chore(docs): updated docs with latest changes

feature(icons): Default color is now currentColor

Icons now inherit the color from their parent as a default

---------

* This puts back the story that allows the user to adjust color/scale of icons.
* Updates the README to be correct with the new way icons are exported/imported.
* Also sets the defaultColor of all icons to the svg css property, currentColor, so that icons inherit color from their parent.

### DIFF
--- a/src/Icon/README.md
+++ b/src/Icon/README.md
@@ -1,21 +1,13 @@
 # Icon
 
-The Icon component is actually an object which holds all Icon components. 
-Trying to render `<Icon />` will result in an error, but for example, `<Icon.ArrowBack />` would not.
+All icons are rendered individually by their name, i.e. **`ArrowBack`** or**`Expand`**. 
 
 ## Usage
-    import { Icon } from '@retailmenot/anchor';
+   
+    import { ArrowBack } from '@retailmenot/anchor';
     
     const example = props => (
-        <Icon.ArrowBack />
-    );
-    
-Alternatively, each icon can be imported individually.
-    
-    import { IconArrowBack } from '@retailmenot/anchor';
-    
-    const example = props => (
-        <IconArrowBack />
+        <ArrowBack />
     );    
     
 ## Settings/Props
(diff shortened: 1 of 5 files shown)


src/Icon/utils.ts
```diff
@@ -19,7 +19,7 @@ export const Scale = {
     xxl: 48,
 };
 
-export const DefaultColor = '#222';
+export const DefaultColor = 'currentColor';
```
bjohnsonrmn, line 22: I either haven't ever seen `currentColor` or forgot about it. It's fun 🎠 

Is it possible for `currentColor` to be null?

brandonlilly (approved):
I see no issues

src/Icon/README.md
```diff
@@ -1,21 +1,13 @@
 # Icon
 
-The Icon component is actually an object which holds all Icon components. 
-Trying to render `<Icon />` will result in an error, but for example, `<Icon.ArrowBack />` would not.
+All icons are rendered individually by their name, i.e. **`ArrowBack`** or**`Expand`**. 
 
 ## Usage
-    import { Icon } from '@retailmenot/anchor';
```
bjohnsonrmn, line 7: Are we discouraging importing the icons like this? 
`import * as Icon from '@retailmenot/anchor';`

If we want people to import the icons like that, then let's include it in the examples.

src/Icon/utils.ts
zero0Halo: I don't believe so. Currently there's no global svg color fill-rule set, yet using `currentColor` does not produce anything odd. It appears to default to the body's color.

src/Icon/README.md
zero0Halo: Importing the icons in this way will import ALL of the icons into a bundle, I believe. We don't have an example of that since that's what we're trying to avoid. We're doing it here so that we can show all of them.

bjohnsonrmn (approved):

PR #45, opened 2019-05-06
feat(Button): add button component

brandonlilly:

---------

Adds a Button component


.storybook/config.js
```diff
@@ -11,10 +11,10 @@ import { addReadme } from 'storybook-readme';
 addDecorator(withInfo({
   source:false,
 }));
-addDecorator(addReadme);
+// addDecorator(addReadme);
 addDecorator(withA11y);
 addDecorator(withKnobs);
-addDecorator(withViewport);
+// addDecorator(withViewport);
```
brandonlilly: I was having an issue with the story container being pushed to the left out of view. Removing this fixed it. Need to come up with another solution

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
+    /* Active styles */
+    ${({ forceActive, disabled, revealed, buttonStyles }: StyledButtonProps) =>
+        !disabled &&
+        !revealed &&
+        buttonStyles.active &&
+        css`
+            &:active {
+                ${buttonStyles.active}
+            }
+            ${forceActive && buttonStyles.active}
```
brandonlilly: Is there any way to make this functionality's code (`forceActive`, `forceFocus`, `forceHover`) dev only?

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
+            buttonStyles={buttonStyles}
+            {...props}
+        >
+            {(height < 3 || width < 3) && (
+                <StyledHitbox buttonHeight={height} buttonWidth={width} />
+            )}
+            {Icon && <Icon scale={iconScale} />}
+            {children}
+            {flip && (
+                <React.Fragment>
```
brandonlilly: Should probably make the flip it's own component

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
+            height={height}
+            icon={Icon}
+            iconOnly={iconOnly}
+            circular={circular}
+            variant={variant}
+            borderRadius={borderRadius}
+            buttonStyles={buttonStyles}
+            {...props}
+        >
+            {(height < 3 || width < 3) && (
```
brandonlilly, line 542: If the width was set to below 3rem, but the button ended up being wider (because wide contents) then a horizontal hitbox could be added to the button when it isn't needed.

In practice, the only kosher button that is below 3 rem is the icon-only button, which isn't at risk of being expanded past its defined width.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
 `;
 
 export const Button = ({
     className,
+    flip = false,
+    variant = 'primary',
+    size = 'large',
+    colorTheme,
+    reverse,
+    circular,
```
brandonlilly, line 460: What do we prefer, `circular` or `rounded`?

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
+    variant: Variant;
+    size: ButtonSize;
 }
 
-const ButtonColorStyles: ButtonStylesGroup = {
-    primary: css`
-        border: solid thin ${colors.savvyCyan.base};
-        background-color: ${colors.savvyCyan.base};
-        color: ${colors.white.base};
+const themeDefaults = {
```
brandonlilly, line 57: I don't love this implementation for colors (See [here](https://github.com/RetailMeNot/anchor/pull/45/files#diff-2e2fd1afd183724a048e4d14f7ea5e3fR580)). It maps really well for the primary button, but not all of the variants and reverse variants fit well into this paradigm, with only 1 or 2 of the hues actually being used or some of the hues not being applied cleanly.

For instance, `minimal - primary` only makes use of savvyCyan.dark, but `minimal - secondary` (which is not included in this PR) replaces savvyCyan **dark** with charcoal **light** and also includes a hover color (charcoal.base).

I do like that we aren't totally coupled to our own themes (you can provide your own {base, light, dark}).

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,619 @@ import * as React from 'react';
...
 import styled, { css, FlattenSimpleInterpolation } from 'styled-components';
-import { darken } from 'polished';
+import { transparentize } from 'polished';
 // THEME
 import { colors, fonts, sizes } from '../theme';
+import { Colors } from '../Typography/Typography.component';
+
+type Variant = 'primary' | 'outline' | 'minimal';
+
+type ButtonSize = 'xlarge' | 'large' | 'small' | 'xsmall';
```
brandonlilly: In other places (e.g. icons) we use `xl`, `lg`, `md`, `sm`, `xs`. Should we switch to use those or include them as aliases? Also noting again that `medium` is not currently used.

src/Button/__snapshots__/Button.component.spec.tsx.snap
```diff
@@ -1,8 +1,2779 @@
...
+  -moz-osx-font-smoothing: grayscale;
+  border: solid thin #0998D6;
+  background-color: #0998D6;
+  color: #FFFFFF;
+  padding: 0 1.5rem;
+  font-size: 1rem;
+  height: 3rem;
+  min-width: 12.5rem;
+  width: 3rem;
+  min-width: initial;
```
brandonlilly: We can make `min-width` optionally defined instead of overriding with `initial`.

src/Button/__snapshots__/Button.component.spec.tsx.snap
```diff
@@ -1,8 +1,2779 @@
...
+.c1 {
+  display: inline-block;
+  height: 1.5rem;
+  width: 1.5rem;
+  line-height: 0;
+}
+
+<button
+  className="anchor-button c0"
+  height={3}
```
brandonlilly: Avoid making `height` and `size` included as a dom attribute.

sharkparty (changes requested):
Will review once checks are passing.

zero0Halo (commented):
I'm having a hard time following exactly how all the pieces of this puzzle fit together because it's all one very large module. Could it potentially be broken down into smaller chunks/imports?

sharkparty (dismissed):

sharkparty:
Just a behavioral question: the focus state as defined means that the outline/border persists on click even during onMouseOut: 

May-07-2019 09-20-35

Can you confirm with UX that this is their desired behavior?

.storybook/config.js
sharkparty: Have you tried manually setting the size of your story's container element? I've had to do this.

src/Button/Button.component.tsx
sharkparty: Yeah, I went with all the minimal versions in most of my components. The trouble with creating globals—to your point—is that not all elements have all options (i.e. omitting medium). I would like this to be "consistent" in that it uses the abbreviated format, though.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
 // VENDOR
 import classNames from 'classnames';
 import styled, { css, FlattenSimpleInterpolation } from 'styled-components';
-import { darken } from 'polished';
+import { transparentize } from 'polished';
 // THEME
 import { colors, fonts, sizes } from '../theme';
+import { Colors } from '../Typography/Typography.component';
```
sharkparty: I should probably have defined this in the theme import and not in typography. Can we move that to the theme and import it in Typography and Button?

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
+
+type Variant = 'primary' | 'outline' | 'minimal';
+
+type ButtonSize = 'xlarge' | 'large' | 'small' | 'xsmall';
+
+interface Theme {
```
sharkparty: I refer to this in typography as a hue as it's not a color, but a treatment of a color. How does one get a dark, savvy cyan? Or is that not a permutation we support?

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
     disabled?: boolean;
+    revealed?: boolean;
     showSpinner?: boolean;
     className?: string;
 
+    flip?: boolean;
+    colorTheme?: Theme | Colors;
+    minWidth?: string;
+    block?: boolean;
+    icon?: any;
```
sharkparty, line 29: We should export a const from icons with all the names of the icons. This has become a pretty common sticking point in other places. Stop by my desk and I'll show you how to generate that in like 2s.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
+    circular?: boolean;
+    reverse?: boolean;
+
+    forceHover?: boolean;
+    forceFocus?: boolean;
+    forceActive?: boolean;
```
sharkparty, line 36: I get these for a storybook story, but I wonder if we'll ever need these for any use-case other than that? If not, maybe there's another way we can implement this.

src/Button/Button.component.tsx
brandonlilly: In this case, at least, we have the option (if we want) of shifting the names xlarge -> large and large -> medium, so that medium is the default.

src/Button/Button.component.tsx
brandonlilly: I agree (my above comment:)

>Is there any way to make this functionality's code (forceActive, forceFocus, forceHover) dev only?

sharkparty: Hah, missed that. What about using something as primitive as classes?

src/Button/Button.component.tsx
sharkparty: 🤔

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
+                active: css`
+                    border: solid thin ${colors.white.base};
+                    background-color: ${colors.white.base};
+                `,
+                focus: css`
+                    border: solid thin ${colors.white.base};
+                    background-color: ${colors.white.base};
+                `,
+                focusOutline: css`
+                    box-shadow: 0 0 0 3px ${colors.white.base};
```
sharkparty: `3px` === `0.1875rem` unfortunately long value but ¯\_(ツ)_/¯ ...

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
+                    background-color: transparent;
+                    color: ${base};
+                `,
+                hover: css`
+                    border: solid thin ${base};
+                    background-color: ${base};
+                    color: ${colors.charcoal.light};
+                `,
+                focusOutline: css`
+                    box-shadow: 0 0 0 2px
```
sharkparty, line 192: px => rem

sharkparty: etc. below with the pixels.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,602 @@ import * as React from 'react';
...
+    variant?: Variant;
+    size?: ButtonSize;
 }
 
-interface ButtonStylesGroup {
-    [key: string]: FlattenSimpleInterpolation;
+interface StyledButtonProps extends ButtonProps {
+    iconOnly?: boolean;
+    height: number;
+    borderRadius: string;
```
sharkparty, line 45: Do we want to limit this in any way?

brandonlilly: It's not a prop exposed on the Button itself, so its only used internally

src/Button/Button.component.tsx
brandonlilly: All of the border sizes in `sizes.theme.ts` are defined in `px`, so I guessed that maybe we didn't want borders/outlines to scale with rems.

Want me to change `sizes.theme` to rems as well? And if so do we want border-radius to use rems?

src/Button/Button.component.tsx
sharkparty: circular

src/Button/__snapshots__/Button.component.spec.tsx.snap
sharkparty: Optionally defined is a better approach IMO.

src/Button/Button.component.tsx
sharkparty: As this evolves it could be helpful as a mechanism for communicating the intent of this param to the next dev, but certainly not a requirement.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,580 @@ import * as React from 'react';
...
+    children,
+    icon: Icon,
     ...props
-}: ButtonProps): React.ReactElement<any> => (
-    <StyledButton
-        className={classNames('anchor-button', className)}
-        {...props}
-    />
-);
+}: ButtonProps): React.ReactElement<any> => {
```
sharkparty: `}: ButtonProps): React.ReactElement<any> => {`
can be
`}: ButtonProps): React.ReactElement<ButtonProps> => {`

src/Button/__snapshots__/Button.component.spec.tsx.snap
brandonlilly: Already done!

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,580 @@ import * as React from 'react';
...
+        : 'md';
+
+    const height = dimensions[size].height;
+    const borderRadius = circular
+        ? `${height / 2}rem`
+        : sizes.border.radius.base;
+    const width = iconOnly ? dimensions[size].height : dimensions[size].width;
+
+    // find theme from string, otherwise use defaults
+    if (typeof colorTheme === 'string') {
```
sharkparty: This is the kind of polymorphism I'm usually against in JS. Not asking you to change this, but I find any kind of type-check in JS to be unreliable. In most of the cases we use this, this'll work, though. What happens if I add a random string, though? Do we want that to crash?

brandonlilly:
> Just a behavioral question: the focus state as defined means that the outline/border persists on click even during onMouseOut:
> 
> May-07-2019 09-20-35
> 
> Can you confirm with UX that this is their desired behavior?

I met with Jen and it turns out that she actually wants the focus outline when tabbing through the buttons. She doesn't really care about having an outline when clicked or anything like that and would prefer it to go away after awhile. She says she's okay with leaving it as is and up to the consumer of the button to defocus.

src/Button/Button.component.tsx
brandonlilly: Yeah I can see that. Currently if they provided a random string it'd just use the defaults.

Which reminds me, there are some props that don't go together. `flip` buttons don't have a `disabled` state, and non `flip` buttons don't have a `revealed` state. `circular` buttons should never have a `flip`. How do we want to handle situations like these? Warning? Crash?

sharkparty: Warnings, but also following up with UX.

src/Button/Button.component.tsx
sharkparty: Worth adding a note in a usage guideline in the docs.

package.json
```diff
@@ -100,6 +100,7 @@
     "enzyme-to-json": "^3.3.4",
     "eslint-plugin-react-hooks": "^1.6.0",
     "jest": "^24.5.0",
+    "jest-styled-components": "^6.3.1",
```
bjohnsonrmn: @brandonlilly just as a heads up, I merged my badge ticket to master yesterday which also adds `jest-styled-components`.

src/Button/Button.component.tsx
brandonlilly: paste list here.

sharkparty: ```type AnchorIcons = 'AddEvent' | 'ArrowBack' | 'ArrowForward' | 'AvatarIcon' | 'AvatarOutline' | 'BarCode' | 'BulletList' | 'Calendar' | 'Camera' | 'Cart' | 'CashBack' | 'Cells' | 'Chat' | 'Check' | 'CheckSmall' | 'ChevronDown' | 'ChevronDownSmall' | 'ChevronLeft' | 'ChevronLeftSmall' | 'ChevronRight' | 'ChevronRightSmall' | 'ChevronUp' | 'ChevronUpSmall' | 'Clock' | 'Close' | 'CloseSmall' | 'CommentMore' | 'CreditCard' | 'CrossHairs' | 'Cut' | 'Disabled' | 'Dislike' | 'Download' | 'Ellipses' | 'EllipsesVertical' | 'Envelope' | 'EnvelopeOpen' | 'Error' | 'Expand' | 'Gear' | 'GiftCard' | 'Hamburger' | 'Heart' | 'HeartOutline' | 'Home' | 'Info' | 'InfoOutline' | 'Laptop' | 'Lightning' | 'Like' | 'ListIcon' | 'Lock' | 'Map' | 'Marker' | 'MarkerOutline' | 'Mobile' | 'News' | 'Pencil' | 'Play' | 'Plus' | 'PlusSmall' | 'Print' | 'Question' | 'QuestionOutline' | 'Refresh' | 'RetailMeNotLogo' | 'SadFace' | 'Search' | 'Share' | 'Sliders' | 'Star' | 'StarHalf' | 'StarOutline' | 'Success' | 'SuccessOutline' | 'Tag' | 'TagAdd' | 'Upload';```

brandonlilly: atm it takes a react element, not the name. e.g.
```jsx
<Button icon={Close} />
```
not
```jsx
<Button icon="Close" />
```
so that the user can potentially provide their own Icons. The button just renders the provided Icon with a `scale` prop.

sharkparty: That's better than my idea.

src/Button/Button.component.tsx
brandonlilly: I thought about this solution for awhile- the issue I'm have with using classes is that the dom state styles to apply with `classes` are a function of a bunch of the button's props (color, circular etc).

I refactored it a little at least so that all of the `force*` css is behind a single `(forceHover || forceFocus || forceActive)` condition.

src/Button/Button.component.tsx
```diff
@@ -3,113 +3,540 @@ import * as React from 'react';
...
+const ButtonColorStyles = ({
+    colorTheme,
+    reverse,
+}: {
+    colorTheme: Theme;
+    reverse?: boolean;
+}): ButtonStylesGroup => {
+    const { base, dark, light } = colorTheme;
+
+    if (reverse) {
```
zero0Halo, line 138: I feel like lines 137 - 281 can largely be abstracted into separate files, making this particular file ~150 lines shorter and easier to read, unless I'm missing why it needs to here.

src/Button/Button.component.tsx
zero0Halo: I don't see a hard reason to keep the props, but I don't think it's completely out of the realm of possibility that someone might want to force any of those states. If anything it provides more programmatic options and that's not a bad thing. As long as it's documented well, I say keep them. That said, add a comment to this effect in the code.

src/Button/Button.component.tsx
zero0Halo: Based on our conversation I say push it out. This method is tightly integrated into the component as it stands.

src/Button/Button.component.tsx
zero0Halo: As I was brutally shot down, never-mind this.

zero0Halo (approved):
LGTM

PR #47, opened 2019-05-09
feat: improve badge component

bjohnsonrmn:

---------

My code improves the Badge component by:
* Adding props to make the size, colors, and positioning configurable
* Adding stories to give examples of options
* Creating snapshot and unit tests
* Filling out the README with info on usage and props

### DIFF
--- a/src/Badge/Badge.component.tsx
+++ b/src/Badge/Badge.component.tsx
@@ -2,15 +2,49 @@
 import * as React from 'react';
 // VENDOR
 import classNames from 'classnames';
-import styled from 'styled-components';
+import styled, { css, FlattenSimpleInterpolation } from 'styled-components';
 // THEME
-import { fonts, colors, sizes, variables } from '../theme';
+import { colors, fonts } from '../theme';
 
-export interface BadgeProps {
-    count?: string | number;
-    className?: string;
+interface BadgeProps {
+    backgroundColor?: string;
+    borderColor?: string;
+    borderColorHover?: string;
     children?: any;
+    className?: string;
+    count?: number;
+    isParentHovered?: boolean;
+    overflowCount?: number;
+    offsetBottom?: number;
+    offsetLeft?: number;
+    showZero?: boolean;
+    size?: 'dot' | 'small' | 'large';
+    standalone?: boolean;
+    textColor?: string;
 }
+
+interface BadgeStylesGroup {
+    [key: string]: FlattenSimpleInterpolation;
+}
+
+const BadgeSizeStyles: BadgeStylesGroup = {
+    dot: css`
+        height: 0.75rem;
+        width: 0.75rem;
+    `,
+    small: css`
+        font-size: 0.625rem;
+        height: 1.25rem;
+        min-width: 1.25rem;
+        padding: 0 0.25rem;
+    `,
+    large: css`
+        font-size: 0.75rem;
+        height: 1.75rem;
+        min-width: 1.75rem;
+        padding: 0 0.5rem;
+    `,
+};
 
 const StyledBadgeContainer = styled.div`
     position: relative;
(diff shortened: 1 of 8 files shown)


sharkparty (changes requested):
Can you make sure the CI passes?

bjohnsonrmn:
@sharkparty yes, let me see if I can get the CI to pass. It seems that it is failing on global branch test coverage.

bjohnsonrmn:
@sharkparty CI is passing 🎉 

I had to clean up a few linting errors in `Badge` and I also somehow picked up a `Button` change that I needed to remove.

src/Badge/Badge.component.tsx
```diff
@@ -2,45 +2,148 @@
...
+    display: flex;
     font-weight: 700;
-    font-size: 0.75rem;
-    padding: 1px ${sizes.padding.xs};
-    border-radius: ${variables.borderRadius};
+    justify-content: center;
+    text-align: center;
+
+    /* Border Radius */
+    ${({ count = 0, size = 'dot' }: BadgeProps) =>
```
bjohnsonrmn: The linter is saying these lines are uncovered:
```
 70,76,86 
```

I'm not sure how to test those. Any ideas?

sharkparty (approved):
LGTM

src/Badge/Badge.component.tsx
sharkparty: Coverage reports all have their anomalies. Likely it'd be more difficult to test this than the value that would be offered by the test. I typically focus on gaining coverage of other uncovered JS files instead of trying to endlessly chase a few lines that I'm pretty confident in. In order to test these you'd need to mock both forks of the logic in your ternary inside of a spec.

src/Badge/Badge.component.tsx
```diff
@@ -2,45 +2,148 @@
...
-        <StyledBadge>{count || '•'}</StyledBadge>
-    </StyledBadgeContainer>
-);
+}: BadgeProps): JSX.Element => {
+    const getDisplayCount = (
+        badgeCount: number,
+        badgeOverflowCount: number
+    ) => {
+        if (badgeCount === 0) {
+            return '';
```
sharkparty, line 121: This was a desired behavior: '•'

sharkparty (changes requested):

src/Badge/Badge.component.tsx
bjohnsonrmn: @sharkparty that makes sense. I think I have tests for the forks of the ternary, just not the assignment of the defaults. Could you remind me what the longhand of this is? 
```
 ${({ count = 0, size = 'dot' }: BadgeProps) =>
```

bjohnsonrmn: @sharkparty that makes sense. I think I have tests for the forks of the ternary, just not the assignment of the defaults. Could you remind me what the longhand of this is?

``` 
${({ count = 0, size = 'dot' }: BadgeProps) =>
```

src/Badge/Badge.component.tsx
bjohnsonrmn: Let me sync with Jen on this. I think she said she didn't want that, though let's make sure we have all the cases covered.

bjohnsonrmn: @sharkparty I talked with Jen and she said she does not want a `•`.  

Our conversation did bring up an uncovered case though. She wants the default to be that the Badge does not show if the `count` is `0`, but she also wants the option to render a Badge with a value of `0`, so I'm planning to add a `showZero` prop that defaults to `false` like Ant does.

sharkparty: Engineering wants the ability to use that • as a value. So long as that is possible, I'm cool with this implementation.

sharkparty:
@bjohnsonrmn I just rebased this against master. Make sure to pull before changing anything.

src/Badge/Badge.component.tsx
bjohnsonrmn: Sure, we can make that possible. Can you tell me more about when we would use `•` instead of a number? I'm wondering if I need to connect it to a certain state of the component or just display it correctly if it is passed in via the `count` prop.

sharkparty: It was a request made by Andy during the UI Guild that I added. I think it's how we show that badge on RMN core if we don't have a count available but do know that there is a notification. I really didn't push back on it bc it was pretty trivial to add.

bjohnsonrmn: Hey Kam, I just chatted with Andy and he said that using the dot size of the badge would work for the use case you mentioned. Are you ok with that solution?

sharkparty: Absolutely. He’s the stakeholder, so if he opts into something else then that’s great!

sharkparty (approved):
LGTM x 2

sharkparty (approved):
👍

sharkparty (approved):

PR #49, opened 2019-05-15
feat(Modal): add modal component

brandonlilly:

---------

Adds a Modal component


src/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,271 @@
...
+            </Typography>
+        )}
+        {children}
+    </StyledHeader>
+);
+
+// Modal.Footer
+// ------------------------------------------------------------------------------------------------------------------
+
+const StyledFooter = styled.div<ModalFooterProps>`
```
sharkparty: We should isolate Header, Footer, & Content into separate files. I'm seeing a lot of conflation and I'd like to break some of these things up more granularly like Autocomplete, Menu, & List who have nested components.

sharkparty (changes requested):
Made a minor request for a change.

src/Modal/README.md
```diff
@@ -0,0 +1 @@
+# Modal
```
zero0Halo, line 1: Need a bit more here

brandonlilly: Should I focus on this or on the doc site md?

src/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+// VENDOR
+import * as StyledReactModal from 'styled-react-modal';
+import styled, { css } from 'styled-components';
+import classnames from 'classnames';
+import { opacify } from 'polished';
+
+export type ModalSize = 'lg' | 'sm';
+
+const Sizes = {
```
zero0Halo, line 21: I noticed that wherever these int's were being used, ultimately they were being interpolated  to rem's. Is there a reason not to specify the unit type here?

brandonlilly: No real reason in this PR. In the button PR I ended up switching from strings to ints a few times so that I could more easily do math on the values, so this time I just started with ints.

src/Modal/Footer/Footer.spec.tsx
```diff
@@ -0,0 +1,20 @@
...
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+
+// STYLED COMPONENTS
+/* tslint:disable no-import-side-effect*/
+import 'jest-styled-components';
+/* tslint:enable */
+
+// COMPONENT
+import { Modal } from '..';
```
sharkparty, line 11: Are you getting a circular import because of this? I had to be verbose with my sub-component `index.ts` imports and use the component file instead.

brandonlilly: Nah, because it's in the spec file, which isn't imported by anything.

src/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+const CustomModalBackground = styled(StyledReactModal.BaseModalBackground)`
+    ${({ oversized }: BaseModalBackgroundProps) =>
+        oversized &&
+        css`
+            align-items: initial;
+        `}
+    overflow-y: auto;
+    background-color: ${({ opacity = 0.6 }: BaseModalBackgroundProps) =>
+        opacify(opacity, 'rgba(0,0,0,0)')};
```
sharkparty: How does this work? It's taking a zero valued alpha of black and then changing the opacity? Why not just `rgb(0, 0, 0)`?

src/Modal/README.md
sharkparty: This can be a follow-up from our chat. I think there will be several doc site tickets.

sharkparty (approved):
Made a few nitpicks, but this LGTM

src/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+    ${StyledFooter} {
+        min-height: ${({ size = defaultSize }: ModalProps) =>
+            Sizes[size].footerHeight}rem;
+    }
+`;
+
+interface ModalProps extends StyledReactModal.ModalProps {
+    size?: ModalSize;
+    background?: string;
+    oversized?: boolean;
```
brandonlilly: I added `oversized` for cases where the modal is very long and needs to not go off the screen, have margins from the screen edge and be scrollable. I believe I can handle this case automatically however if I add a wrapper around the modal.

sharkparty:
FYI-You can overwrite the lock file without any regard for what it's replacing. We can chat about this but tl;dr: we don't really want any lock files in this repo.

src/Modal/Modal.component.tsx
brandonlilly: Just kidding, I think I just realized a way to do it automatically even without another dom element.

src/Modal/Modal.component.tsx
brandonlilly: I believe the way `opacify` works it adds additional opacity to something already semi-transparent. It doesn't _set_ the opacity, so rgba(0,0,0) would always remain at opacity 1.

But why don't I just do `rgba(0,0,0,${opacity})`? Derp

sharkparty: I've avoided the docs for polished as I'm on the fence about deprecating it, so I did not know that's how that method worked. Definitely like the leaner approach you outlined, though.

PR #52, opened 2019-05-28
feat(Button): prefixes and suffixes

brandonlilly:

---------

This adds a `prefix` and `suffix` prop to Button and removes the `icon` prop. A button is now `iconOnly` if is now just a `prefix` or `suffix` with no children.


src/Button/Button.component.tsx
```diff
@@ -460,10 +467,12 @@ export const Button = ({
...
     children,
     minWidth,
-    icon: Icon,
+    prefix,
+    suffix,
     ...props
 }: ButtonProps): React.ReactElement<ButtonProps> => {
-    const iconOnly = Icon && React.Children.count(children) === 0;
+    const iconOnly =
+        (prefix ? !suffix : !!suffix) && React.Children.count(children) === 0;
```
brandonlilly, line 475: >`(prefix ? !suffix : !!suffix)`

This is effectively `prefix XOR suffix`

src/utils/cloneWithProps/cloneWithProps.ts
```diff
@@ -0,0 +1,21 @@
...
+    element: React.ReactElement<any>,
+    props: Props = {}
+) => {
+    const { className, ...rest } = props;
+    return React.cloneElement(element, {
+        ...rest,
+        ...element.props,
+        className: classNames(className, element.props.className),
+    });
+};
```
zero0Halo, line 27: I think a comment block explaining why you built such a robust cloning service would be helpful.

src/Button/Button.component.tsx
zero0Halo: Add a comment explaining what's going on here. This borders on too clever. :)

src/Button/Button.component.tsx
```diff
@@ -370,16 +373,20 @@ const StyledButton = styled.button<StyledButtonProps>`
...
 
-    ${({ icon, iconOnly, $size }: StyledButtonProps) =>
-        icon &&
+    ${({ iconOnly, $size }: StyledButtonProps) =>
         !iconOnly &&
         css`
             // Space icon from text
-            & > .anchor-icon {
+            & > .anchor-button-prefix {
                 margin-right: ${$size === 'lg' || $size === 'md'
```
zero0Halo: Since you're comparing against a string value (and in multiple places), might be a good idea to assign the strings to constants to avoid magic numbers. i.e. const LG = 'lg'

src/Button/Button.component.tsx
```diff
@@ -370,16 +373,20 @@ const StyledButton = styled.button<StyledButtonProps>`
...
                     ? 0.5
                     : 0.375}rem;
             }
+            & > .anchor-button-suffix {
+                margin-left: ${$size === 'lg' || $size === 'md'
```
zero0Halo: Other spot w/magic-numbers

src/Button/Button.component.tsx
```diff
@@ -495,8 +504,7 @@ export const Button = ({
     /* tslint:enable no-console */
 
     const iconScale = iconOnly
-        ? size === 'xs' ||
-          (size === 'sm' && (circular || variant === 'minimal'))
+        ? size === 'xs' || (size === 'sm' && variant === 'minimal')
```
zero0Halo, line 507: Magic numbers

src/Button/Button.component.tsx
brandonlilly: I'm not sure this is really necessary because it's enforced by types, which are defined above. If I add consts for all these then I'm duplicating the information already outlined in the ButtonSize type. What do you think?

zero0Halo: It has less to do with that it is enforced by types, and more to do with making a comparison in code, particularly because `'lg'` and `'md'` are used in multiple places. It's a logical expression and IMO, those should always be against a `const` and not an inline string literal (or number).

brandonlilly: Most of this will be reorganized when we switch to using a theme solution, so I'm going to punt on this and keep it in mind when that refactor happens.

sharkparty (approved):
LGTM

PR #54, opened 2019-05-29
feat: Adding CollapseGroup component

zero0Halo:

---------

This component pulls together multiple `Collapse` components together, allowing for many individual props to be set for all of them at once, make them all look like they belong together, and give the option to have accordion functionality.


src/Collapse/Collapse.component.tsx
```diff
@@ -76,10 +79,15 @@ export const Collapse = ({
...
+            className={classNames(
+                'anchor-collapse',
+                open && 'open',
+                theme,
+                className
+            )}
         >
             <button
-                onClick={() => toggleOpen(!open)}
+                onClick={() => (toggleFn ? toggleFn() : toggleOpen(!open))}
```
brandonlilly:
```js
toggleFn || () => toggleOpen(!open)
```
if you prefer

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+    accordion = false,
+    openIndex = 0,
+    theme,
+    openIcon,
+    closeIcon,
+    children,
+    className,
+}: CollapseGroupProps) => {
+    const [index, changeIndex] = React.useState<number>(
+        openIndex === 0 ? 0 : openIndex
```
brandonlilly: isn't this the same thing as
```jsx
React.useState<number>(openIndex)
```
?

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    return (
+        <StyledCollapseGroup
+            className={classNames('anchor-collapse-group', className)}
+        >
+            {
+                // tslint:disable-next-line: max-line-length
+                /** If accordion is true, all Collapse components are cloned in order to control their isOpen prop, and to use the changeIndex function to do so  */
+            }
+            {accordion
```
brandonlilly:
```jsx
{children &&
children.map((child: any, i: number) => {
    return React.cloneElement(child, {
        isOpen: accordion ? i === index : undefined,
        toggleFn: accordion ? () => changeIndex(i) : undefined,
        theme: theme,
        openIcon: openIcon,
        closeIcon: closeIcon,
        key: `collapse-child-${i}`,
    });
})}
```
Perhaps?

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+export const CollapseGroup = ({
+    accordion = false,
+    openIndex = 0,
+    theme,
+    openIcon,
+    closeIcon,
+    children,
+    className,
+}: CollapseGroupProps) => {
+    const [index, changeIndex] = React.useState<number>(
```
brandonlilly: I would rename this to be slightly more descriptive, something like 
```js
const [openIndex, setOpenIndex] = ...
```

src/Collapse/Collapse.component.tsx
```diff
@@ -53,9 +55,10 @@ export const Collapse = ({
     isOpen = false,
     openText = DEFAULT_OPEN_TEXT,
     closeText = DEFAULT_CLOSE_TEXT,
-    openIcon = false,
-    closeIcon = false,
+    openIcon,
+    closeIcon,
     theme = DEFAULT_THEME,
+    toggleFn,
```
brandonlilly: I would rename to something more conventional like `onClick` or `onToggle`.

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+// VENDOR
+import classNames from 'classnames';
+import styled from 'styled-components';
+// COMPONENTS
+import { Themes } from '../Collapse.component';
+
+// The mundane function of CollapseGroup is to remove the bottom border of any Collapse children,
+// save for the last one.
+const StyledCollapseGroup = styled.div`
+    .anchor-collapse:not(:last-child) {
```
brandonlilly: Optional: You can use a reference to styled components directly instead of like `.anchor-collapse`.
```css
${StyledCollapse}:not(:last-child) {
  
  & ${StyledCollapseButton} {
```

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+            {
+                // tslint:disable-next-line: max-line-length
+                /** If accordion is true, all Collapse components are cloned in order to control their isOpen prop, and to use the changeIndex function to do so  */
+            }
+            {accordion
+                ? children &&
+                  children.map((child: any, i: number) => {
+                      return React.cloneElement(child, {
+                          isOpen: i === index,
+                          toggleFn: () => changeIndex(i),
```
brandonlilly: The way `accordion` is implemented currently, it isn't possible to close the currently open collapse so that everything is close. I was surprised by this when I was playing around on storybook.

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
```diff
@@ -0,0 +1,83 @@
+// REACT
+import * as React from 'react';
+// VENDOR
+import classNames from 'classnames';
+import styled from 'styled-components';
+// COMPONENTS
+import { Themes } from '../Collapse.component';
+
+// The mundane function of CollapseGroup is to remove the bottom border of any Collapse children,
```
bjohnsonrmn: Would this be worth a unit test?

src/Collapse/CollapseGroup/CollapseGroup.component.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+    it("should show only one Collapse component's content when accordion is true.", () => {
+        const subject = (
+            <CollapseGroup accordion openIndex={1}>
+                <Collapse>Hello World</Collapse>
+                <Collapse>Hello World</Collapse>
+                <Collapse>Hello World</Collapse>
+            </CollapseGroup>
+        );
+        const wrapper = mount(subject);
+        const toggleButton = wrapper.find('.anchor-collapse-button').first();
```
bjohnsonrmn, line 41: Is this simulating clicking the `Collapse` that is already open? I just want to make sure this is testing the functionality of clicking a different `Collapse` and _then_ making sure there is still only 1 open.

src/Collapse/CollapseGroup/CollapseGroup.component.spec.tsx
```diff
@@ -0,0 +1,56 @@
+// REACT
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+// Vendor
+import { shallow, mount } from 'enzyme';
+// COMPONENT
+import { CollapseGroup } from './CollapseGroup.component';
+import { Collapse } from '../Collapse.component';
+// import { ArrowBack } from '../../Icon';
```
bjohnsonrmn: Perhaps this comment could be removed?

src/Collapse/Collapse.stories.tsx
```diff
@@ -60,36 +82,52 @@ storiesOf('Components/Collapse', module)
...
                 </Collapse>
             </StyledComfortableStory>
         );
+    })
+    .add('CollapseGroup', () => {
+        const content = select('Example Content', [TEXT, 'list'], TEXT);
+        const iconOpenPick = select('iconOpen', Object.keys(Icon), 'ChevronUp');
+        const iconClosePick = select(
+            'iconClose',
+            Object.keys(Icon),
```
bjohnsonrmn, line 104: This is cool ✅

src/Collapse/README.md
```diff
@@ -39,6 +39,47 @@ The component can accept 3 `props`:
...
+* `toggleFn`: Whatever function is passed to this prop will be executed when the user clicks the **`.collapse-button`** element. It will not intrinsically toggle the **`isOpen`** prop.
+<br /><br />     
+
+# CollapseGroup
+
+The **`CollapseGroup`** component combines multiple **`Collapse`** components together, allowing a single component to apply props to all children, unifying their appearance, and adding the ability to have accordion functionality.
+
+## Usage
+   
+    import { Collapse, CollapseGroup } from '@retailmenot/collapse';
```
bjohnsonrmn: Does `anchor` need to be in this import?

bjohnsonrmn (approved):
LGTM. I left some small questions and suggestions.

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
zero0Halo: This is a leftover for when I needed something better than a truthiness test. I'll remove it.

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
zero0Halo: I dig it. I hated having the two child loops.

src/Collapse/Collapse.component.tsx
zero0Halo: I don't like reusing event names like onClick, and I'm not sure onToggle quite fits either since that sounds like an event. toggleFn is a function that controls the toggling of open/close of a Collapse child. It could potentially use something more descriptive, but I'd like it to be more representative of that. Ideas?

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
zero0Halo: I went this route to give an indication that there were a number of classes that a user could use to style the component.

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
zero0Halo: This is a **really** good point. I will work on adding that functionality.

src/Collapse/CollapseGroup/CollapseGroup.component.spec.tsx
zero0Halo: It simulates clicking a closed one. The `openIndex` prop that is passed makes it so the second Collapse child is open, then the simulated click is on the first one which is closed.

src/Collapse/Collapse.component.tsx
brandonlilly: Well... it is an event though. It *isn't* a function that controls the toggling of the collapse, it is really just a function that runs when the button is clicked. It's up to the parent to decide what that function does.

I suggested just `onToggle` initially, but then I remembered https://jxnblk.com/blog/defining-component-apis-in-react/#keep-props-apis-parallel and added `onClick` as my primary suggestion

zero0Halo: I don't agree that it's an event. An event is `onClick`, `onMouseOver` etc., right? What happens when those events happen is an event handler. Maybe that's the answer though, what about `onToggleHandler` or something similar?

src/Collapse/Collapse.component.tsx
```diff
@@ -53,9 +55,10 @@ export const Collapse = ({
     isOpen = false,
     openText = DEFAULT_OPEN_TEXT,
     closeText = DEFAULT_CLOSE_TEXT,
-    openIcon = false,
-    closeIcon = false,
+    openIcon,
```
brandonlilly: The default close and open icons are backwards I think. This isn't related to your PR, just something pre-existing I noticed.

src/Collapse/Collapse.component.tsx
brandonlilly: Well yes, it is an event *handler* for the onClick event. This is typically how react handler props are defined however.
```jsx
<button onClick={handleButtonClick} ></button>
<Collapse onClick={() => toggleThisGuyAndCloseEveryOneElse()} />
```

The reason I suggest this change is actually because I was having difficulty understanding what the CollapseGroup was doing until I realized that the `toggleFn` prop is just the onClick handler. And in the case of the Accordion, the behavior of the Collapse when clicked is to close all other collapses.

zero0Halo: Yep, I think we're in agreement now. I'll change the wording both in `Collapse` and `CollapseGroup` to try and make it more clear that `CollapseGroup` is passing a new handler to `Collapse` to use. Good discussion, thanks!

src/Collapse/CollapseGroup/CollapseGroup.component.tsx
bjohnsonrmn: It took me a bit to figure out how to do this. Here's an example from `Badge` that may help:
https://github.com/RetailMeNot/anchor/blob/master/src/Badge/Badge.component.spec.tsx#L57

and you need `jest-styled-components` (which will also change your snapshots)

zero0Halo: I think that `jest-styled-components` has a hard time interpreting pseudo selectors. I used a fun one, `.anchor-collapse:not(:last-child)` and it can't seem to recognize the style even though in the browser I can see that it's working. This is something I would like to test for, but I'm wondering if it's possible.

bjohnsonrmn: Ah yes, Enzyme might not support pseudo selectors yet. 

Would these work for finding the first and last ones?
```
find('.anchor-collapse').first()
find('.anchor-collapse').last()
```

bjohnsonrmn: If that doesn't work, here is another possibility I found (less elegant but might still be useful):

```
find({ key: 'collapse-child-0' })
find({ key: 'collapse-child-2' }); 
```
from here: https://airbnb.io/enzyme/docs/api/selector.html#4-object-property-selector

zero0Halo: No matter what I did I couldn't get enzyme to play ball with this. I refactored the code to give `Collapse` a `hasBottomBorder` prop which adds a class to hide the border. Seeing as how that's the primary reason to have `CollapseGroup` exist it seemed prudent to have a test that actually checked it.

src/Collapse/Collapse.component.tsx
zero0Halo: Yeah I went back and forth on this. But seeing as how I often make this mistake it probably makes sense to change it.

src/Collapse/Collapse.component.tsx
sharkparty: Sorry to be late to this party, this is a solid discussion.

### RE Event Handling
I think the react team has some opinions on event handling [naming conventions](https://reactjs.org/docs/events.html). It seems that they rely on a `on` prefix denoting that this is an event handler, an event name `on[EventName]`, and then a modifier for meta (think pre/post) event handlers `on[EventName][Modifier]` so for a bubbling-handler `onClickCapture` works. Then, inside the component itself, they use `handle[EventName]` for the internal component method handling the event.

### RE Event-Handling By Reference
I can't find any opinionated naming conventions around this, but the method being described here is a trigger, so what if we all named instance trigger methods with a `trigger` prefix? `triggerCollapse(true)` to open or `triggerCollapse(false)` to close? I'm not dead set on `trigger` so much as I'd like to define a consistent naming convention for these types of events to that in our docs anytime someone sees a keyword they will intuit that it's a trigger-method for controlling a component instance.

sharkparty: Looking more closely at the code and the [readme](https://github.com/RetailMeNot/anchor/pull/54/files#diff-6fdb9d96efcd884c4398b16c8de687ab) (which may not have been updated with this feedback), it doesn't look like this is a trigger. It doesn't allow an external component to trigger events, it's an internal event handler, so I think this is more of a traditional event handler (i.e. `on[EventName]`) than a trigger.

PR #58, opened 2019-06-05
feat(Modal): create theme solution

brandonlilly:

---------

My code does...


brandonlilly:
I made this PR to give a preview into how Styled System might integrate into our app, and as a forum to talking about more concrete details.

src/Modal/Modal.component.tsx
```diff
@@ -18,73 +28,82 @@ import { ModalClose } from './Close';
...
-    margin: ${({ margin = 'auto' }: ModalProps) => margin};
+
+    ${compose(
+        marginStyles,
+        backgroundStyles,
+        colorStyles,
+        borderRadiusStyles
+    )}
+
+    ${size(
```
brandonlilly, line 62: This is using a custom `variant` util I made for handling more complex logic of theme settings.

The provided `variant` function from styled-components and styled-system would require each variant to be css style objects, which can't really handle this use-case.

brandonlilly: Alternatively I could do something like create a util for mapping the active variant properties as props to the component. This could potentially result in some prop name conflicts.

src/Modal/Header/Header.component.tsx
```diff
@@ -15,16 +25,20 @@ export const StyledHeader = styled.div<ModalHeaderProps>`
...
-    color: ${({ color = colors.charcoal.light }) => color};
-    background: ${({ background = colors.white.base }) => background};
+    ${compose(
+        colorStyles,
+        backgroundStyles,
+        borderRadiusStyles
+    )}
+
+    && {
+        ${paddingStyles}
```
sharkparty, line 40: What's this doing?

brandonlilly: I separated out the padding like this because the padding is normally defined by the parent component (Modal). This  increases the specificity so that this css takes precedence if it exists.

sharkparty: Got it. Neat.

src/Modal/Modal.component.tsx
```diff
@@ -18,73 +26,90 @@ import { ModalClose } from './Close';
...
+});
 
 const StyledModal = StyledReactModal.default.styled`
     position: relative;
-    width: ${({ width, size = defaultSize }: ModalProps) =>
-        width || `${Sizes[size].width}rem`};
-    height: ${({ height = '42.375rem' }: ModalProps) => height};
-    margin: ${({ margin = 'auto' }: ModalProps) => margin};
+
+    ${compose(
```
brandonlilly, line 55: This could be written alternatively like
```tsx
${marginStyles}
${backgroundStyles}
${colorStyles}
${borderRadiusStyles}
```

I wrote it with `compose` because Styled System recommends it for better performance of multiple styles https://styled-system.com/api#compose

src/Modal/Modal.component.tsx
```diff
@@ -4,12 +4,20 @@ import * as React from 'react';
 // VENDOR
 import * as StyledReactModal from 'styled-react-modal';
 import styled, { css } from 'styled-components';
+import {
+    compose,
+    variant,
+    borderRadius as borderRadiusStyles,
```
brandonlilly, line 10: I aliased these to avoid shadow-naming tslint errors, due to the prop variables given defaults having the same names.

src/Modal/Modal.component.tsx
sharkparty: Nah, I saw the docs on this and like the approach you took.

src/Modal/Modal.component.tsx
sharkparty: I found this out when trying to delete them locally. Not a bad way to avoid the linter errors. We could modify the rules, but I like this better.

src/Modal/Content/Content.component.tsx
```diff
@@ -9,12 +20,25 @@ export const StyledContent = styled.div<ModalContentProps>`
...
+        ${paddingStyles}
+    }
+
     order: 0;
     flex-grow: 1;
 `;
 
-export interface ModalContentProps {
+export interface ModalContentProps
+    extends ColorProps,
```
bjohnsonrmn, line 37: This took me a bit to wrap my head around, but I like it ✅

src/Modal/Footer/__snapshots__/Footer.spec.tsx.snap
```diff
@@ -22,7 +23,6 @@ exports[`Component: Modal.Footer should render 1`] = `
   -ms-flex-order: 1;
   order: 1;
   justify-self: flex-end;
-  background: transparent;
```
bjohnsonrmn, line 25: Was removing this intentional?

src/Modal/Modal.component.tsx
```diff
@@ -4,87 +4,110 @@ import * as React from 'react';
...
 import { ModalFooter, StyledFooter } from './Footer';
 import { ModalClose } from './Close';
 
-export type ModalSize = 'lg' | 'sm';
-
-const Sizes = {
-    lg: { width: 50, contentPadding: 2.5, footerHeight: 6.5 },
-    sm: { width: 25, contentPadding: 2, footerHeight: 5.5 },
+export const MODAL_THEME_KEY = 'modal';
+export const MODAL_THEME = {
```
bjohnsonrmn, line 28: This theme object is helpful for understanding modal variations 👍

src/Modal/Modal.component.tsx
```diff
@@ -4,87 +4,110 @@ import * as React from 'react';
...
+    )}
+
+    ${size(
+        ({ contentPadding = 0, footerHeight, width, height }) => css`
+            height: ${({ height: propHeight }: ModalProps) =>
+                propHeight || `${height}rem`};
+            width: ${({ width: propWidth }: ModalProps) =>
+                propWidth || `${width}rem`};
+
+            // The Modal Content component only gets its bottom padding if
```
bjohnsonrmn, line 69: I wonder if this complexity would be easier to manage at a higher level. Perhaps a separate component that is just the `ModalContent` that could be used in cases where we know we need the full padding. Or perhaps even another theme option that could offer different frill levels? Not sure if either would be easier, but just tossing it out there.

src/theme/system.ts
```diff
@@ -0,0 +1,80 @@
+// ANCHOR
+import { radii } from './sizes.theme';
+
+// VENDOR
+import {
+    get,
+    system,
+    TLengthStyledSystem,
```
bjohnsonrmn, line 8: What is `TLengthStyledSystem`?

src/theme/system.ts
```diff
@@ -0,0 +1,80 @@
...
+    BackgroundProps,
+    padding,
+    PaddingProps,
+    margin,
+    MarginProps,
+    display,
+    DisplayProps,
+} from 'styled-system';
+
+export const variant = <T extends {}>({
```
bjohnsonrmn, line 27: Is this still vendor code all the way down to lines 43?

brandonlilly: Just the imports

src/theme/system.ts
brandonlilly: Using it to define the Type Defs for BorderRadiusProps

PR #60, opened 2019-06-11
fix: Collapse listed in exports

zero0Halo:

---------

The Collapse/CollapseGroup component(s) were not listed among the exports in the rollup config. As such the entire component was included with all other components. This fix will correctly put the component(s) into its own file rather than into index.js.

This will require an upgrade to version .30 as anyone using Anchor will be importing Collapse whether they want to or not.


brandonlilly (approved):
Perhaps this should also be added as a step in the PR template

zero0Halo:
> Perhaps this should also be added as a step in the PR template

Good idea. I added it.

PR #64, opened 2019-06-20
feat: Added sidenav to doc site

zero0Halo:
chore: Upgraded anchor to .29

feat(Pages): Added stubs for all components that exist in Anchor

fix: Moved side navigation from Page into SideNavigation component

feat(context): Added context to track user navigation

feat: Adding in react hook to grab location

feat: Sidenavigation opens correct Collapse element

chore: added Interface's for Section and Link

feat: Added styles to active GatsbyLink components

refactor: Removed navigation context

I decided that the context was no longer necessary as I switched approaches for auto-opening sections in the sidenav

chore(cleanup): Added more comments

chore: Ran prettier

---------

This adds a working side navigation to the doc site. Mostly every page is a stub so there is considerable work ahead, but this is the first step.


docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
```diff
@@ -1,130 +1,87 @@
...
-            padding-left: ${sizes.padding.lg};
-        }
+const StyledCollapseGroup = styled(CollapseGroup)`
+    li a {
+        text-decoration: none;
 
-        &:last-of-type {
-            padding-bottom: ${sizes.padding.xl};
+        &.active {
+            background-color: #e3eef6;
```
bjohnsonrmn, line 20: Perhaps it would be a good idea to import the Anchor theme and use a variable here instead of a hex?

docs/src/components/Navigation/SideNavigation/sections.ts
```diff
@@ -0,0 +1,94 @@
+export const sections = [
```
bjohnsonrmn, line 1: Consider adding links to the theme and utilities pages here. Utilities is all placeholder at at the moment, so we could add it later once it's useful, but theme already has the color swatch lists so could potentially already add some value (and I have already directed one person to it who was looking for it but couldn't find it)

bjohnsonrmn (commented):
LGTM. I had a few minor suggestions, but it's cool to see the side nav coming together to provide the structure for us to start filling in the docs--and you were able to use an Anchor component to implement 💯 Good work!

docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
```diff
@@ -1,130 +1,87 @@
...
+    // I'm using PureComponent in order to use a constructor since ComponentWillMount is being deprecated.
+    // I need to get the index of the current section in order to pass that value to CollapseGroup.
+    // This is what makes the correct Collapse component open when navigating the site.
+    constructor(props: object) {
+        super(props);
 
-                &:hover {
-                    background-color: ${colors.silver.dark};
-                }
+        sections.map((section: SectionProperties, i: number) => {
```
brandonlilly: This isn't returning anything so it should be a `forEach` instead of a `map`

brandonlilly (approved):

docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
zero0Halo: Although this is a good idea, ironically I don't believe the doc site conforms to the colors of RMN's theme. I will dig through it to see if I am wrong (which I'd be really happy to be in this instance).

zero0Halo: These colors aren't in our theme, so I added a TODO comment noting that when we have our Theme Provider in place that they should be moved into it.

PR #68, opened 2019-07-01
fix: No longer need window.location for sidenav 

zero0Halo:
feat: New HOC, AddLocation. Gives any component access to the router's location prop

fix: Using ^, no longer have to use window.location which was making the build fail

fix: updated all paths so that activeClassName correctly is fired for sidenav

chore: Moved logic from constructor into componentDidMount

chore: prettier updates

---------

This should take care of two issues. 1) No longer need to access window.location from the SideNavigation component. That's good because it was causing the build to fail.  2) No longer have the initialization code in the constructor, it can exist within the componentDidMount lifecycle method.

### DIFF
--- a/docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
+++ b/docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
@@ -42,16 +42,13 @@ class SideNavigation extends React.PureComponent<SideNavigationProps> {
     };
 
     componentWillMount() {
-        let mainOpenIndex = 0;
+        const { pathname } = this.props.location;
 
-        sections.forEach((section: SectionProperties, i: number) => {
+        const mainOpenIndex = sections.reduce((openIndex: number, section: SectionProperties, index: number) => {
             const { pattern } = section;
-            const { pathname } = this.props.location;
             // Compares the current url with the path associated to a section and gets its index if it matches.
-            if (pattern.length && pathname.includes(pattern)) {
-                mainOpenIndex = i;
-            }
-        });
+            return (pattern.length && pathname.includes(pattern)) ? index : openIndex;
+          }, 0);
 
         this.setState({mainOpenIndex});
     }
--- a/docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
+++ b/docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
@@ -7,6 +7,7 @@ import styled from '@xstyled/styled-components';
 import { Collapse, CollapseGroup } from '@retailmenot/anchor';
 // TODO: Change the config to allow ts extensions. This works, but tsconfig is having a snit.
 import { sections } from './sections';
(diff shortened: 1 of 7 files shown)


src/Modal/Modal.component.tsx
```diff
@@ -49,7 +49,9 @@ const StyledModal = StyledReactModal.default.styled`
     // Footer (respectively) are also used.
     ${StyledContent} {
         padding: ${({ size = defaultSize }: ModalProps) =>
-            `4rem ${Sizes[size].contentPadding}rem 0 ${Sizes[size].contentPadding}rem`};
+            `4rem ${Sizes[size].contentPadding}rem 0 ${
```
sharkparty: Just curious: what version of prettier are you on locally?

zero0Halo: `prettier@1.17.1`

docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
```diff
@@ -31,32 +36,33 @@ interface LinkProperties {
...
+    componentWillMount() {
+        let mainOpenIndex = 0;
 
-    componentDidMount(): void {
         sections.forEach((section: SectionProperties, i: number) => {
             const { pattern } = section;
-            const { pathname } = window.location;
+            const { pathname } = this.props.location;
             // Compares the current url with the path associated to a section and gets its index if it matches.
             if (pattern.length > 0 && pathname.includes(pattern)) {
```
sharkparty: `pattern.length` and `pattern.length > 0` resolve the same way

zero0Halo:
The side nav is broken right now (visually). Hold off on any further reviewing until I figure out what happened.

zero0Halo:
Side nav is no longer visually borked

docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
```diff
@@ -31,32 +36,33 @@ interface LinkProperties {
...
             }
         });
+
+        this.setState({mainOpenIndex});
     }
 
     render() {
-        return (
+        const { mainOpenIndex } = this.state;
+
```
brandonlilly, line 58: I like the functional approach:
```js
const { pathname } = this.props.location;

const mainOpenIndex = sections.reduce((openIndex: number, section: SectionProperties, index: number) => {
  const { pattern } = section;
  // Compares the current url with the path associated to a section and gets its index if it matches.
  return (pattern.length && pathname.includes(pattern)) ? index : openIndex
}, 0);

this.setState({ mainOpenIndex });
```

brandonlilly (approved):

docs/src/components/Navigation/SideNavigation/SideNavigation.component.tsx
zero0Halo: I dig it. Updated.

PR #75, opened 2019-07-19
fix: toggle double onChange, modal border radius and modal exports

brandonlilly:

---------

Stops Toggle from triggering onChange twice
Fixes Modal border radius (didn't have any)
Exports ModalProvider and BaseModalBackground from Modal

### DIFF
--- a/src/Form/Toggle/Toggle.component.tsx
+++ b/src/Form/Toggle/Toggle.component.tsx
@@ -139,6 +139,7 @@ export const Toggle = forwardRef(
             trackHeight = '0.75rem',
             trackWidth = '2rem',
             disabled,
+            onChange,
             ...props
         }: ToggleProps,
         ref: React.RefObject<any>
@@ -164,6 +165,7 @@ export const Toggle = forwardRef(
             <HiddenInput
                 type="checkbox"
                 id={id}
+                onChange={onChange}
                 checked={checked}
                 disabled={disabled}
                 ref={ref}
--- a/src/Modal/Modal.component.tsx
+++ b/src/Modal/Modal.component.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 // VENDOR
 import * as StyledReactModal from 'styled-react-modal';
 import styled, { css } from '@xstyled/styled-components';
+import { th } from '@xstyled/system';
 import classnames from 'classnames';
 // COMPONENTS
 import { ModalHeader, StyledHeader } from './Header';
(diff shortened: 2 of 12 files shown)


src/Form/Toggle/Toggle.component.tsx
```diff
@@ -164,6 +165,7 @@ export const Toggle = forwardRef(
             <HiddenInput
                 type="checkbox"
                 id={id}
+                onChange={onChange}
```
brandonlilly, line 168: onChange was being spread across both the `label` and the `input` causing the onChange handler to be called twice.

src/Modal/Modal.component.tsx
```diff
@@ -31,10 +32,12 @@ const StyledModal = StyledReactModal.default.styled`
     flex-direction: column;
     justify-content: space-between;
 
-    border-radius: modal;
-    ${({ background = 'neutrals.white.base' }: ModalProps) =>
-        css({ backgroundColor: background })};
-    ${({ color }: ModalProps) => css({ color })}
+    ${({ background = 'neutrals.white.base', color }: ModalProps) =>
+        css({
+            background: th.color(background),
```
brandonlilly, line 37: `linear-gradient` didn't work for `backgroundColor`, so I switched it back to `background` but manually run it through the theme color.

src/Modal/Modal.component.tsx
```diff
@@ -31,10 +32,12 @@ const StyledModal = StyledReactModal.default.styled`
...
+            color,
+            borderRadius: 'modal',
```
brandonlilly, line 39: This component is defined like so `StyledReactModal.default.styled` so it doesn't run through the xstyled interpolation. I moved the borderRadius prop to the xstyled `css` function so that it is parsed as expected.

src/theme/index.ts
```diff
@@ -24,4 +24,7 @@ export const RootTheme = {
     radii: { none: '0', base: '4px', modal: '8px', circular: '1000px' },
     [BUTTON_KEY]: BUTTON_THEME,
     colors: ColorsTheme,
+    breakpoints: {
```
brandonlilly, line 27: Just adding `xs` to the breakpoints by default so that if someone extends our theme they don't have to do it themselves. The xs breakpoint is just the default behavior (no media query).

src/Form/Toggle/Toggle.component.tsx
sharkparty: 👍

src/theme/index.ts
sharkparty: 👍

sharkparty (approved):
LGTM; bump your version if you need a publish.

PR #76, opened 2019-07-19
feat: added content for Typography component for doc site

zero0Halo:
feat: added readme text

chore: added details for typography's props

chore: Added a section for default styles

chore: started making the props section look more like autocomplete's api section

chore: made the props section the api section, all bulleted props are now in a table

---------

This adds documentation for the Typography component in the doc site. In addition, it's my first attempt at documenting a component that I did not myself build. Comments are very much encouraged on the accuracy of my understanding.


docs/src/pages/components/typography.mdx
```diff
@@ -1 +1,186 @@
...
+* `font-family: Avenir Next, Segoe UI, Helvetica Neue, Helvetica, Roboto,sans-serif;`
+* `font-size: 1rem;`
+* `line-height: 1.5rem;`
+* `text-align: inherit;`
+* `text-transform: none;`
+* `color: inherit;`
+
+<br />
+
+#### Inline Code Editor
```
sharkparty, line 28: We should add a descriptor in here about the reason we have this notion of `scale`. It's (1) our way of mitigating the fact that semantic HTML tagging doesn't always get the same text styles and we needed a mechanism for handling that and (2) we needed a consistent way to render common fonts regardless of their enclosing tags for CTAs.

sharkparty (changes requested):
Adding an addendum for `scale` vs `size`/`weight`/`line-height`

docs/src/pages/components/typography.mdx
zero0Halo: Updated description: "This value directly affects the size and lineHeight props directly, ensuring
                that there is no overlap of text between lines no matter what value is input.
                Often, semantic HTML tagging doesn't always get the same text styles and this helps
                alleviate that problem. It also assists in rendering common fonts regardless of 
                their enclosing tags."

zero0Halo: I've added a paragraph to this effect.

sharkparty (approved):
LGTM

PR #79, opened 2019-07-29
chore: adding base props

sharkparty:

---------

My code allows users to apply any React props to our components (i.e. `style={{ padding: '1rem' }}`).


src/Menu/Menu.component.tsx
```diff
@@ -31,7 +31,7 @@ const StyledMenu = styled('nav')`
   //overflow-y: hidden;
   //overflow-x: auto;
 
-  ${({ bg = th.color('primary.base') }: MenuProps) =>
+  ${({ bg = th.color('primary.base') }) =>
```
brandonlilly: not related to the pr, but the `th.color`s here and below are redundant

brandonlilly: also unrelated: this reminds me that when using style utils from `@xstyled/system` you get aliases for the same prop, e.g. `mb=` and `marginBottom=`, we may need to be mindful of how we provide defaults for these props.

i.e. it would not be ideal if the user has to override the specific alias we used to get the marginBottom they want.

brandonlilly:
I haven't fully thought through the logical consequences quite yet, but the one thing I'm wary about is potential overlap between `React.HTMLAttributes<HTMLDivElement>` props and other props we may want to use, e.g. 
```
import { color as colorStyles, ColorProps } from '@xstyled/system`

interface ComponentProps extends ColorProps, React.HTMLAttributes<HTMLDivElement> {

}
```

Afaik, they both define the `color` prop. Does the order you extend in matter? Might they interfere? Would we potentially fool people into thinking we have an interface that they don't (e.g. Button now has a color prop, but it wouldn't do anything iirc)

src/Menu/Menu.component.tsx
sharkparty: Nice catch on these. I lost the forrest through the trees. 

Also, this is something I've had in the back of my mind but I really don't have an elegant solution. We're balancing the tension of flexibility or convention and I don't have a "best" idea yet. If you do, I'm all ears.

sharkparty (commented):
They override and extend compatibly (for the most part). Both will get passed to the child component and get used; however, in some cases collisions may occur with how the props are being used. I don't think it's a gate until we understand more about where the collisions lie. From there, we can define our own `interface` that's more rigid. I didn't want to do so in this PR until we had the usage data to back an approach.

src/Menu/Menu.component.tsx
brandonlilly: Don't have a best solution either. I did a little testing and it seems that the order in which things are defined on the component is what determines what will take precedence (which surprises me). e.g.
```
// will have 5rem margin-bottom
<Box mb="1rem" marginBottom="5rem" />

// will have 1rem margin-bottom
<Box marginBottom="5rem" mb="1rem" />
```

which means if a component is defined like this:

```
// props is spread after marginBottom
const Box = ({ marginBottom = "-2rem", ...props }) => (
  <StyledBox marginBottom={marginBottom} {...props} />
);

// margin-bottom will be 5rem
<Box mb="5rem" />
```
it'll work "properly", but if its defined like this:
```
// props is spread before marginBottom
const Box = ({ marginBottom = "-2rem", ...props }) => (
  <StyledBox {...props} marginBottom={marginBottom} />
);

// margin-bottom will be -2rem
<Box mb="5rem" />
```
it may not

(If I took the correct lessons from the tests I did)

brandonlilly (approved):
>From there, we can define our own interface that's more rigid. I didn't want to do so in this PR until we had the usage data to back an approach.

Agreed. LGTM

PR #80, opened 2019-07-30
feat(Checkbox): add checkbox component

brandonlilly:

---------

Adds a Checkbox component

### DIFF
--- a/src/Form/Checkbox/Checkbox.component.tsx
+++ b/src/Form/Checkbox/Checkbox.component.tsx
@@ -8,7 +8,7 @@ import { compose } from '@xstyled/system';
 import classnames from 'classnames';
 
 // ANCHOR
-import { Check } from '../..';
+import { Check } from '../../Icon';
 import { StyledIcon } from '../../Icon/utils';
 import {
     space as spaceStyles,
@@ -38,6 +38,7 @@ const StyledCheckbox = styled('label')<CheckboxProps>`
 
     user-select: none;
     border: dark;
+    // todo: may want to read from a theme value here
     border-radius: 2px;
 
     ${compose(
--- a/src/Form/Checkbox/Checkbox.component.tsx
+++ b/src/Form/Checkbox/Checkbox.component.tsx
@@ -1,0 +1,123 @@
+// REACT
+import * as React from 'react';
+const { forwardRef } = React;
+
+// VENDOR
+import styled, { css } from '@xstyled/styled-components';
+import { compose } from '@xstyled/system';
+import classnames from 'classnames';
+
+// ANCHOR
+import { Check } from '../..';
+import { StyledIcon } from '../../Icon/utils';
+import {
+    space as spaceStyles,
+    SpaceProps,
+    color as colorStyles,
+    ColorProps,
+    backgroundColor as backgroundColorStyles,
+    BackgroundColorProps,
+} from '../../theme/system.theme';
+
+const HiddenInput = styled('input')<CheckboxProps>`
+    visibility: hidden;
+    opacity: 0;
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+`;
+
+const StyledCheckbox = styled('label')<CheckboxProps>`
+    box-sizing: border-box;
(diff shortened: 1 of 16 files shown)


src/Form/Checkbox/Checkbox.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+import * as React from 'react';
+const { forwardRef } = React;
+
+// VENDOR
+import styled, { css } from '@xstyled/styled-components';
+import { compose } from '@xstyled/system';
+import classnames from 'classnames';
+
+// ANCHOR
+import { Check } from '../..';
```
sharkparty: Does this give you a circular reference error on build?

src/Form/Checkbox/Checkbox.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+const StyledCheckbox = styled('label')<CheckboxProps>`
+    box-sizing: border-box;
+    position: relative;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+
+    user-select: none;
+    border: dark;
+    border-radius: 2px;
```
sharkparty, line 42: I'm guessing we don't want to codify this as a `small`?

src/Form/Checkbox/Checkbox.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+`;
+
+interface CheckboxProps extends SpaceProps, ColorProps, BackgroundColorProps {
+    className?: string;
+    id?: string;
+    htmlFor?: string;
+
+    checked?: boolean;
+    disabled?: boolean;
+    name?: string;
```
sharkparty, line 74: I don't know if this should be optional.

src/Form/Checkbox/Checkbox.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+    ${StyledIcon} {
+        transition: opacity 100ms ease-in-out;
+        ${({ checked }) =>
+            css({
+                opacity: checked ? 1 : 0,
+            })}
+    }
+`;
+
+interface CheckboxProps extends SpaceProps, ColorProps, BackgroundColorProps {
```
sharkparty, line 67: Given the event handlers, you may not be able to use `React.HTMLAttributes<HTMLInputElement>`, but can you verify that? I know AutoComplete has too many custom handlers to be compatible with that interface.

sharkparty (changes requested):
Extremely minor feedback. This is mostly solid.

src/Form/Checkbox/Checkbox.component.tsx
brandonlilly: Maybe we want to. Not sure how snowflake it is

src/Form/Checkbox/Checkbox.component.tsx
brandonlilly: Woops

src/Form/Checkbox/Checkbox.component.tsx
sharkparty: Let's leave this as is, but can you add a TODO calling this out?

src/Form/Checkbox/Checkbox.component.tsx
brandonlilly: Wasn't compatible, but how about:
```
    inputProps?: React.HTMLAttributes<HTMLInputElement>;
```
?

sharkparty: This won't allow event binding on the parent checkbox, but it does solve a part of the issue. It's ok if this is omitted for now. I need to figure out a FormControl interface that allows us more flexibility.

src/Form/Checkbox/Checkbox.component.tsx
brandonlilly: I started switching and noticed that I have `ControlLabel`s automatically propagate it's `name` down to the control you give it. So if `name` was required, you'd have to specify the `name` for both anyway.

sharkparty (approved):

PR #84, opened 2019-08-01
docs: Adds Hero to doc site

zero0Halo:
fix: Rolled back ability to pass a title to CodePreview

feat: CodePreview has default title's displayed, and can be hidden via prop

docs: Updated several mdx files to take advantage of the above change

docs: Added API and a bigger example of the Hero

---------

Adds Hero documentation to the doc site. Also rolls back a feature introduced in a previous build, being able to pass a title as a prop to `CodePreview`. This turned out not to work because props passed by triple back-tick in an mdx file are limited to a single no-space variable.  Instead, a default title is always displayed, though can be hidden by passing the `hideTitle` prop.

### DIFF
--- a/docs/src/components/CodePreview/CodePreview.component.tsx
+++ b/docs/src/components/CodePreview/CodePreview.component.tsx
@@ -12,6 +12,13 @@
     ```tsx live
         <AddEvent />
     ```
+
+    A default title of "Inline Code Editor" or "Example" is displayed by default depending on if the
+    'live' prop is passed. To hide it, pass the hideTitle prop.
+
+    ```tsx hideTitle
+        <AddEvent />
+    ```
 */
 
 // REACT
@@ -30,7 +37,12 @@ import Component from '@reach/component-component';
 import Highlight, { defaultProps } from 'prism-react-renderer';
 // tslint:disable-next-line: no-submodule-imports
 import github from 'prism-react-renderer/themes/github';
-import { AutoComplete, Collapse, CollapseGroup } from '@retailmenot/anchor';
+import {
+    AutoComplete,
+    Collapse,
+    CollapseGroup,
+    Typography,
+} from '@retailmenot/anchor';
 // COMPONENTS
 import * as Anchor from '../../../../src';
 // THEME
@@ -117,11 +129,16 @@ const StyledLivePreview = styled(LivePreview)<PreProps>`
     border-bottom-right-radius: 0.25rem;
 `;
 
+const StyledCodeBlock = styled('pre')`
+    padding: 1.25rem;
+    background-color: ${colors.silver.base};
+`;
+
 interface CodePreviewProps {
     children?: any;
     className?: string;
     live?: boolean;
-    title?: string;
+    hideTitle?: boolean;
 }
 
 const scope = {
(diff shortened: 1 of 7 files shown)


docs/src/components/CodePreview/CodePreview.component.tsx
```diff
@@ -166,29 +176,43 @@ export const CodePreview = ({
...
+                    tokens,
+                    getLineProps,
+                    getTokenProps,
+                }) => (
+                    <pre
+                        className={className}
+                        style={{ ...style, padding: '20px' }}
+                    >
+                        {tokens.map((line, i) => (
+                            <div key={i} {...getLineProps({ line, key: i })}>
```
brandonlilly, line 205: `key` shouldn't be the index value of a map function -sharkparty

docs/src/components/CodePreview/CodePreview.component.tsx
```diff
@@ -166,29 +176,43 @@ export const CodePreview = ({
...
+                {({
+                    className,
+                    style,
+                    tokens,
+                    getLineProps,
+                    getTokenProps,
+                }) => (
+                    <pre
+                        className={className}
+                        style={{ ...style, padding: '20px' }}
```
brandonlilly: inline-styles ? -kam

zero0Halo: Totally going to try and get rid of that. Just depends on what I can get away with in modifying MDX's code.

docs/src/pages/components/hero.mdx
```diff
@@ -1,4 +1,199 @@
...
+    const { Title, Subtitle } = Hero;
+```
+
+Depending on how your environment is set up, you could directly access these sub-components directly
+in your jsx as shown below. This example uses the `colors` export from **Anchor** which is the basis for <Link to="/theme">themes</Link>,
+as well as the <Link to="/components/button/">`Button`</Link> component.
+
+```tsx live
+    <Hero background={colors.grapePurchase.base}>
+        <Typography tag="div" p="4">
```
brandonlilly: There should be a way of adding padding to the Hero itself but unfortunately I overlooked that. So instead of adding a Typography with padding, I'd set the `minHeight` of the `Hero` to, say, `12.5rem`, and then in the future we can change it.

docs/src/components/CodePreview/CodePreview.component.tsx
zero0Halo: Spoke w/Kam on this. There isn't a great way to fix this particular issue and as it's static content it shouldn't matter much. Point taken for future usage.

docs/src/components/CodePreview/CodePreview.component.tsx
zero0Halo: Made a styled component for this piece.

docs/src/pages/components/hero.mdx
zero0Halo: Done.

docs/src/components/CodePreview/CodePreview.component.tsx
```diff
@@ -117,11 +129,16 @@ const StyledLivePreview = styled(LivePreview)<PreProps>`
...
 
+const StyledCodeBlock = styled('pre')`
+    padding: 1.25rem;
+    background-color: ${colors.silver.base};
+`;
+
 interface CodePreviewProps {
     children?: any;
     className?: string;
     live?: boolean;
```
sharkparty, line 140: I'm wondering if this could be an enum of types instead of having a prop for each type?

sharkparty (approved):
LGTM with some minor feedback

PR #94, opened 2019-08-12
fix(build): update xstyled to 1.10.0

zero0Halo:

---------

@xstyled v1.8.4 introduced a bug where one of the modules from their `utils` repo was not being pulled into their `styled-components` repo correctly. Consequently the build process was failing. The bug was later fixed in version 1.10.0.

This branch also adds a number of components to the `typings.d.ts` file under `@retailmenot/anchor`.

### DIFF
--- a/docs/src/components/Layout/Page/Page.component.tsx
+++ b/docs/src/components/Layout/Page/Page.component.tsx
@@ -4,7 +4,8 @@ import * as React from 'react';
 import classNames from 'classnames';
 import { MDXProvider } from '@mdx-js/tag';
 import { AutoComplete } from '@retailmenot/anchor';
-import styled, { css, ThemeProvider } from '@xstyled/styled-components';
+import styled, { ThemeProvider } from '@xstyled/styled-components';
+import { css } from 'styled-components';
 // COMPONENTS
 import { fonts, NormalizeCSS } from '../../../../../src/theme';
 import { RootTheme } from '../../../../../src/theme';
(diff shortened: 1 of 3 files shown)


docs/src/components/Layout/Page/Page.component.tsx
```diff
@@ -4,7 +4,8 @@ import * as React from 'react';
 import classNames from 'classnames';
 import { MDXProvider } from '@mdx-js/tag';
 import { AutoComplete } from '@retailmenot/anchor';
-import styled, { css, ThemeProvider } from '@xstyled/styled-components';
+import styled, { ThemeProvider } from '@xstyled/styled-components';
+import { css } from 'styled-components';
```
sharkparty, line 8: This seems like a step backward.

zero0Halo: Real problem is that I'm not 100% sure what is causing the issue. I have a gut feeling this may be it, but until I can hit the deploy endpoint where it's failing I can't tell. This is more of a step to figuring out the cause. If this does fix it, then the next step is figuring out how to get `css` from `xstyled` to work.

sharkparty: Can't you run the CI scripts locally minus the deploy?

sharkparty: https://github.com/RetailMeNot/anchor/blob/master/Dockerfile

`docker build -t client .`
`docker run client`

sharkparty (approved):
LGTM, please add comment

PR #96, opened 2019-08-15
feat(dropdown): refactored dropdown and added mobile support

sharkparty:

---------

My code fixes some bugs with the DropDown component, adds the ability to position it more granularly, and adds mobile support.

### DIFF
--- a/src/DropDown/Arrow/Arrow.component.tsx
+++ b/src/DropDown/Arrow/Arrow.component.tsx
@@ -11,6 +11,7 @@ const ArrowBase = css`
     height: 0;
 `;
 
+// TODO: add shadow rotate(45deg)
 const CaretHeight = '0.4rem';
 const CaretIndent = '0.75rem';
 const CaretSize = '0.5rem';
--- a/src/DropDown/Arrow/Arrow.component.tsx
+++ b/src/DropDown/Arrow/Arrow.component.tsx
@@ -1,0 +1,74 @@
+// REACT
+import * as React from 'react';
+// VENDOR
+import styled, { css } from '@xstyled/styled-components';
+// COMPONENTS
+import { Position } from '../DropDown.component';
+
+const ArrowBase = css`
+    position: absolute;
+    width: 0;
+    height: 0;
+`;
+
+const CaretHeight = '0.4rem';
+const CaretIndent = '0.75rem';
+const CaretSize = '0.5rem';
+
+export const ArrowUp = styled('div')`
+    ${ArrowBase};
+    top: -${CaretHeight};
+    left: ${CaretIndent};
+    border-left: ${CaretSize} solid transparent;
+    border-right: ${CaretSize} solid transparent;
+    border-bottom: ${CaretSize} solid white;
+`;
+
+export const ArrowDown = styled('div')`
+    ${ArrowBase};
+    bottom: -${CaretHeight};
+    left: ${CaretIndent};
+    border-left: ${CaretSize} solid transparent;
+    border-right: ${CaretSize} solid transparent;
+    border-top: ${CaretSize} solid white;
+`;
+export const ArrowRight = styled('div')`
+    ${ArrowBase};
+    top: ${CaretIndent};
+    right: -${CaretHeight};
+    border-top: ${CaretSize} solid transparent;
+    border-bottom: ${CaretSize} solid transparent;
+    border-left: ${CaretSize} solid white;
+`;
+
+export const ArrowLeft = styled('div')`
+    ${ArrowBase};
+    top: ${CaretIndent};
+    left: -${CaretHeight};
+    border-top: ${CaretSize} solid transparent;
+    border-bottom: ${CaretSize} solid transparent;
+    border-right: ${CaretSize} solid white;
+`;
+
+export const renderArrow = (position: Position, isShown?: boolean) => {
+    let arrowElement;
+    if (isShown) {
+        switch (position) {
+            case 'bottom':
+                arrowElement = <ArrowUp />;
+                break;
+            case 'top':
+                arrowElement = <ArrowDown />;
+                break;
+            case 'right':
+                arrowElement = <ArrowLeft />;
+                break;
+            case 'left':
+                arrowElement = <ArrowRight />;
+                break;
+            default:
+                break;
+        }
+    }
+    return arrowElement;
+};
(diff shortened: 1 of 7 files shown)


zero0Halo:
General comment, knob all the props.

src/DropDown/DropDown.component.tsx
```diff
@@ -60,48 +81,122 @@ const StyledDropDownContainer = styled(DropDownContainer)<
...
     /* Defined Position */
-    top: ${({ position: { top } = {} }) =>
-        top ? `${top}px !important` : 'inherit'};
-    left: ${({ position: { left } = {} }) =>
-        left ? `${left}px !important` : '0'};
-    right: ${({ position: { right } = {} }) =>
-        right ? `${right}px !important` : 'inherit'};
-    bottom: ${({ position: { bottom } = {} }) =>
-        bottom ? `${bottom}px !important` : 'inherit'};
+    ${({ position, height, width, offset = '0.5rem' }) => {
```
zero0Halo, line 96: `offset` doesn't appear to be used

src/DropDown/DropDown.component.tsx
```diff
@@ -60,48 +81,122 @@ const StyledDropDownContainer = styled(DropDownContainer)<
...
+const ArrowLeft = styled('div')`
+    ${ArrowBase};
+    top: ${CaretIndent};
+    left: -${CaretHeight};
+    border-top: ${CaretSize} solid transparent;
+    border-bottom: ${CaretSize} solid transparent;
+    border-right: ${CaretSize} solid white;
+`;
+
+const renderArrow = (position: Position, isShown?: boolean) => {
```
zero0Halo: There is a lot of logic/syntax just for the arrow, plus here its parts are assembled into a functional component. Perhaps move to its own file?

zero0Halo (commented):
Line breaks are a cheap and easy way to separate logic. A lot of the code here is very dense, it could benefit from some space, especially because in some places you use them and other's you don't.

src/DropDown/DropDown.component.tsx
sharkparty: See line 20. This isn't something I wanted to ship with MVP, but this will likely be an optimization we need.

zero0Halo: Got it.

src/DropDown/DropDown.stories.tsx
```diff
@@ -36,28 +52,67 @@ storiesOf('Components/DropDown', module)
...
+                        </Col>
+                    </Row>
+                    <Row>
+                        <Col offset={{ xs: 3, sm: 3, md: 3, lg: 3 }} lg={4}>
+                            <div>
+                                <Typography tag="h1">DropDown Top</Typography>
+                                <DropDown
+                                    overlay={<MockList />}
+                                    position="top"
+                                    trigger="click"
```
brandonlilly, line 109: This one is `click` and all the rest are hover. Made me wonder in the story if it was broken for a couple seconds. Idk if this was intentional.

src/DropDown/DropDown.component.tsx
```diff
@@ -181,15 +257,23 @@ export class DropDown extends React.Component<DropDownProps> {
             <StyledDropDown
                 ref={this.dropDownReference}
                 className={classNames('anchor-drop-down', className)}
+                position={position || 'bottom'}
```
brandonlilly, line 197: add default? `{ position = 'bottom' } = this.props`

zero0Halo (approved):
I still think you could use a few more line breaks, but otherwise LGTM.

src/DropDown/Arrow/Arrow.component.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const CaretHeight = '0.4rem';
+const CaretIndent = '0.75rem';
+const CaretSize = '0.5rem';
+
+export const ArrowUp = styled('div')`
+    ${ArrowBase};
+    top: -${CaretHeight};
+    left: ${CaretIndent};
+    border-left: ${CaretSize} solid transparent;
```
brandonlilly, line 23: It looks a little bit odd to not have a shadow on the Arrow. If you use the `transform: rotate(45deg)` method then you can give it a shadow and a border if you like, though you'd have to layer the elements just right so that the shadows don't overlap improperly.

src/DropDown/Arrow/Arrow.component.tsx
```diff
@@ -0,0 +1,74 @@
...
+    border-right: ${CaretSize} solid transparent;
+    border-bottom: ${CaretSize} solid white;
```
brandonlilly, line 25: If the background of the dropdown can be modified, then it'll have to get passed on down to the Arrow to match.

src/DropDown/DropDown.component.tsx
```diff
@@ -60,48 +84,54 @@ const StyledDropDownContainer = styled(DropDownContainer)<
 >`
     min-width: 10rem;
     position: absolute;
-    padding: 0.25rem 0;
+    padding: 0.25rem auto;
     z-index: 1;
     background-color: ${th.color('neutrals.white.base')};
```
brandonlilly, line 90: Would be nice to get props for some of these.

brandonlilly (approved):
Looks great. Nothing I suggested needs to be MVP

src/DropDown/DropDown.stories.tsx
sharkparty: It is. I use the storybook to demo across all devices and supported browsers with all permutations of the component that present engineering challenges.

src/DropDown/Arrow/Arrow.component.tsx
sharkparty: see other comment on arrow.

src/DropDown/Arrow/Arrow.component.tsx
sharkparty: This is a good callout. I'll add a todo for this and the shadow note. I honestly couldn't figure that bit out, so thanks!

src/DropDown/DropDown.component.tsx
sharkparty: It would, but it present a ton of challenges making it that composable and I don't think we'll need that yet. 80/20 argument, which is super weak...

PR #100, opened 2019-08-19
feat(Tooltip): added tooltip

sharkparty:

---------

My code adds a Tooltip component to Anchor.

### DIFF
--- a/src/Tooltip/Tooltip.component.tsx
+++ b/src/Tooltip/Tooltip.component.tsx
@@ -20,11 +20,11 @@ export type Position =
     | 'leftStart';
 
 interface TooltipContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+    content: string;
     position?: Position;
-    width?: string;
+    maxWidth?: string;
     background?: string;
     color?: string;
-    content?: any;
     wrap?: boolean;
 }
 
@@ -50,6 +50,7 @@ interface TooltipElementProps {
     wrap?: boolean;
     background?: string;
     color?: string;
+    maxWidth?: string;
 }
 
 const tooltipMargin = 8;
@@ -166,7 +167,9 @@ const TooltipElement = styled('div')<TooltipElementProps>`
     transition: opacity 250ms ease-in-out;
 
     ${({ wrap = true }) =>
-        wrap ? css({ whiteSpace: 'normal' }) : css({ whiteSpace: 'nowrap' })}
+        wrap ? css({ whiteSpace: 'normal' }) : css({ whiteSpace: 'nowrap' })};
+
+    ${({ maxWidth = 'auto' }) => css({ width: maxWidth })};
 
     ${({ position, height, width, toolTipHeight, toolTipWidth }) =>
         positionVariants(position, height, width, toolTipHeight, toolTipWidth)};
@@ -193,6 +196,7 @@ export class Tooltip extends React.Component<
     componentDidMount(): void {
         const { current: toolTipContainer } = this.tooltipContainerRef;
         const { current: toolTip } = this.tooltipRef;
+        // TODO: instead of setting state, what about using the render fxn?
         if (toolTipContainer) {
             this.setState({
                 height: toolTipContainer.clientHeight,
@@ -216,19 +220,22 @@ export class Tooltip extends React.Component<
             wrap,
             background,
             color,
+            maxWidth,
             ...props
         } = this.props;
         const { height, width, toolTipHeight, toolTipWidth } = this.state;
         return (
             <ToolTipContainer
+                content={content}
                 onMouseEnter={() => this.setState({ hidden: false })}
                 onMouseLeave={() => this.setState({ hidden: true })}
                 ref={this.tooltipContainerRef}
-                className={classNames('anchor-badge', className)}
+                className={classNames('anchor-tooltip', className)}
                 {...props}
             >
                 {children}
                 <TooltipElement
+                    className="anchor-tooltip-element"
                     ref={this.tooltipRef}
                     position={position || 'topEnd'}
                     height={height}
@@ -239,6 +246,7 @@ export class Tooltip extends React.Component<
                     wrap={wrap}
                     background={background}
                     color={color}
+                    maxWidth={maxWidth}
                     style={{ opacity: this.state.hidden ? 0 : 1 }}
                 />
             </ToolTipContainer>
--- a/src/Tooltip/Tooltip.component.tsx
+++ b/src/Tooltip/Tooltip.component.tsx
@@ -79,13 +79,6 @@ const positionVariants = (
                     toolTipWidth / 2}px, -${toolTipHeight +
                     tooltipMargin}px, 0)`,
             });
-        case 'topEnd':
-            return css({
-                right: 0,
-                top: 0,
-                transform: `translate3d(0, -${toolTipHeight +
-                    tooltipMargin}px, 0)`,
-            });
         case 'rightStart':
             return css({
                 left: 0,
@@ -146,11 +139,13 @@ const positionVariants = (
                 transform: `translate3d(-${toolTipWidth +
                     tooltipMargin}px, 0, 0)`,
             });
+        case 'topEnd':
         default:
             return css({
                 right: 0,
                 top: 0,
-                transform: `translate3d(0, -${height + tooltipMargin}px, 0)`,
+                transform: `translate3d(0, -${toolTipHeight +
+                    tooltipMargin}px, 0)`,
             });
     }
 };
--- a/src/Tooltip/Tooltip.component.tsx
+++ b/src/Tooltip/Tooltip.component.tsx
@@ -173,11 +173,8 @@ const TooltipElement = styled('div')<TooltipElementProps>`
         visibility: visible;
     }
 
-    ${({ wrap = true }) =>
-        wrap ? css({ whiteSpace: 'normal' }) : css({ whiteSpace: 'nowrap' })};
-
+    ${({ wrap = true }) => css({ whiteSpace: wrap ? 'normal' : 'nowrap' })};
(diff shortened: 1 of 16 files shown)


src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,255 @@
...
+                    height={height}
+                    width={width}
+                    toolTipHeight={toolTipHeight}
+                    toolTipWidth={toolTipWidth}
+                    children={content}
+                    wrap={wrap}
+                    background={background}
+                    color={color}
+                    maxWidth={maxWidth}
+                    style={{ opacity: this.state.hidden ? 0 : 1 }}
```
zero0Halo: Since we're avoiding inline styles, you could instead add the following to `TooltipElement`...
```
    &.active {
        opacity: 1;
        visibility:visible;
    }
```

And change line 238 to:

`className={classNames('anchor-tooltip-element', !this.state.hidden && 'active')}`

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,255 @@
...
+    position: absolute;
+    box-sizing: border-box;
+    ${({ background = 'rgba(0, 0, 0, 0.8)', color = 'white' }) =>
+        css({ background, color })};
+    border-radius: base;
+    font-family: ${th('typography.fontFamily')};
+    font-size: 0.8rem;
+    padding: 0.5rem;
+    opacity: 0;
+    transition: opacity 250ms ease-in-out;
```
zero0Halo, line 164: Because you're only affecting opacity, it's possible to still hover over the tooltip and have it be displayed since its hit box is still in the DOM. If you change 167 & 168, that'll be a step to fixing that. Then refer to my comment on removing inline styles for the next piece.

```
transition: opacity 250ms ease-in-out, visibility 250ms ease-in-out;
visibility:hidden;
```

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,255 @@
...
+    | 'rightStart'
+    | 'right'
+    | 'rightEnd'
+    | 'bottomEnd'
+    | 'bottom'
+    | 'bottomStart'
+    | 'leftEnd'
+    | 'left'
+    | 'leftStart';
+
```
zero0Halo, line 21: I don't envy you having to come up with these names. At first I had no clue what they meant until I played around with the knobs, then it made perfect sense. We should consider reusing these names for all components that have some type of positioning scheme to be consistent.

zero0Halo (commented):
Please (at least) copy the content of your `README` to the doc site. If you need me to format it, make a ticket.

src/Tooltip/Tooltip.component.tsx
sharkparty: I want to preserve the animation, so this won't solve that.

src/Tooltip/Tooltip.component.tsx
sharkparty: I agree. I'll be abstracting and applying this logic to DropDown.

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+    border-radius: base;
+    font-family: ${th('typography.fontFamily')};
+    font-size: 0.8rem;
+    padding: 0.5rem;
+    opacity: 0;
+    visibility: hidden;
+    transition: opacity 250ms ease-in-out;
+
+    &.active {
+        opacity: 1;
```
brandonlilly, line 167: We could add a `delay` prop, so that the tooltip doesn't necessarily appear immediately.

Would be easy to implement, I think, just add a `transition-delay` style to `&.active { }` so that it has a delayed entrance, but no delay when it hides.

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+};
+
+const TooltipElement = styled('div')<TooltipElementProps>`
+    position: absolute;
+    box-sizing: border-box;
+    ${({ background = 'rgba(0, 0, 0, 0.8)', color = 'white' }) =>
+        css({ background, color })};
+    border-radius: base;
+    font-family: ${th('typography.fontFamily')};
+    font-size: 0.8rem;
```
brandonlilly, line 160: `font-size: 0.8rem;`?

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+    visibility: hidden;
+    transition: opacity 250ms ease-in-out;
+
+    &.active {
+        opacity: 1;
+        visibility: visible;
+    }
+
+    ${({ wrap = true }) =>
+        wrap ? css({ whiteSpace: 'normal' }) : css({ whiteSpace: 'nowrap' })};
```
brandonlilly:
```js
css({ whiteSpace: wrap ? 'normal' : 'nowrap' })
```
perhaps?

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+    ${({ maxWidth = 'auto' }) => css({ width: maxWidth })};
```
brandonlilly, line 172: applying `maxWidth` to `width` style?

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+        HTMLDivElement
+    > = React.createRef();
+
+    componentDidMount(): void {
+        const { current: toolTipContainer } = this.tooltipContainerRef;
+        const { current: toolTip } = this.tooltipRef;
+        // TODO: instead of setting state, what about using the render fxn?
+        if (toolTipContainer) {
+            this.setState({
+                height: toolTipContainer.clientHeight,
```
brandonlilly, line 201: No listeners for resizing events or sommat?

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+    width: number;
+    height: number;
+    toolTipHeight: number;
+    toolTipWidth: number;
+    hidden: boolean;
+}
+
+const ToolTipContainer = styled('div')<TooltipContainerProps>`
+    position: relative;
+    display: inline-block;
```
brandonlilly: We may want to make this a prop, because its a wrapper over their content. They may want it to play differently in the DOM than inline-block at times.

src/Tooltip/Tooltip.component.tsx
```diff
@@ -0,0 +1,262 @@
...
+                left: 0,
+                top: 0,
+                transform: `translate3d(-${toolTipWidth +
+                    tooltipMargin}px, 0, 0)`,
+            });
+        default:
+            return css({
+                right: 0,
+                top: 0,
+                transform: `translate3d(0, -${height + tooltipMargin}px, 0)`,
```
brandonlilly: Seems odd that the default positioning can't be explicitly opted into.

src/Tooltip/Tooltip.component.tsx
sharkparty: 👍

src/Tooltip/Tooltip.component.tsx
sharkparty: I think that's outside of MVP. delayStart and delayHide are an evolution IMO

src/Tooltip/Tooltip.component.tsx
sharkparty: 👍

src/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    .addParameters({
+        readme: {
+            sidebar: README,
+        },
+    })
+    .add('Default', () => (
+        <StyledStory>
+            <ThemeProvider theme={RootTheme}>
+                <>
+                    <Typography tag="h1">Top Right</Typography>
```
brandonlilly: Perhaps a more general heading, since this changes with the knobs

brandonlilly:
Looks great in storybook 👍

src/Tooltip/Tooltip.component.tsx
sharkparty: `maxWidth` is a prop for defining the tooltip width. Width is a prop that we use for passing the width of the rendered tooltip around. I was going to change the width prop, but maxWidth seems more semantic.

src/Tooltip/Tooltip.component.tsx
sharkparty: Nope. This isn't a dropdown; it's a tooltip.

src/Tooltip/Tooltip.component.tsx
sharkparty: It's the same as `topEnd`, for now.

src/Tooltip/Tooltip.stories.tsx
sharkparty: 👍

sharkparty:
If I need to make a docs entry, then I'll finish the entire docs page for this comp. I don't think that's what we discussed.

src/Tooltip/Tooltip.component.tsx
brandonlilly: I thought it was `topEnd` at first but I looked closely and topEnd is `toolTipHeight + tooltipMargin` and this is `height + tooltipMargin`;

Maybe we should be explicit about the default and add `position = 'topEnd'` to the this.prop defaults.

sharkparty: That's likely an error on my part.

sharkparty: Fixed. I moved the `topEnd` case to the bottom of the switch.

brandonlilly (approved):

zero0Halo (approved):

PR #109, opened 2019-08-28
feat(dropdown) abstract positioning variants, add trigger for both click and hover

brandonlilly:

---------

My code does...

### DIFF
--- a/src/DropDown/DropDown.stories.tsx
+++ b/src/DropDown/DropDown.stories.tsx
@@ -32,6 +32,7 @@ const MockList = () => (
     <List>
         {listItemArray.map(({ key, label }: any) => (
             <ListItem
+                background='transparent'
                 key={key}
                 onClick={() => {
                     /* tslint:disable no-console */
@@ -163,7 +164,8 @@ storiesOf('Components/DropDown', module)
                                             text('background', '') || undefined
                                         }
                                         border={text('border', '') || undefined}
-                                        indent={text('indent', '') || undefined}
+                                        arrowIndent={text('arrowIndent', '') || undefined}
+                                        arrowSize={text('arrowSize', '') || undefined}
                                         spacing={
                                             spacing
                                                 ? parseInt(spacing)
--- a/src/DropDown/DropDown.stories.tsx
+++ b/src/DropDown/DropDown.stories.tsx
@@ -6,15 +6,15 @@ import { storiesOf } from '@storybook/react';
 import styled, { ThemeProvider } from '@xstyled/styled-components';
 import { text, select, boolean } from '@storybook/addon-knobs';
 // COMPONENTS
-import { DropDown, Position } from './DropDown.component';
+import { DropDown, Position, Trigger } from './DropDown.component';
 import { Button } from '../Button';
 import { List, Item as ListItem } from '../List';
 import { Typography } from '../Typography';
 import { Container, Row, Col } from '../Grid';
+// THEME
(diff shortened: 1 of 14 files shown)


src/Tooltip/Tooltip.component.tsx
```diff
@@ -4,20 +4,9 @@ import * as React from 'react';
...
-    | 'rightEnd'
-    | 'bottomEnd'
-    | 'bottom'
-    | 'bottomStart'
-    | 'leftEnd'
-    | 'left'
-    | 'leftStart';
+// ANCHOR
+import { positionVariants, Position } from '../utils/position/position';
+export { Position } from '../utils/position/position';
```
zero0Halo, line 9: Any reason why `Position` is being exported here? I noticed it was also being exported from `DropDown` too.

src/DropDown/DropDown.stories.tsx
```diff
@@ -127,20 +129,49 @@ storiesOf('Components/DropDown', module)
...
+                                        text('arrowIndent', '') || undefined
+                                    }
+                                    arrowSize={
+                                        text('arrowSize', '') || undefined
+                                    }
+                                    spacing={text('spacing', '') || undefined}
                                 >
-                                    <a>{text('Dropdown Label', 'Dropdown')}</a>
+                                    <Button>
+                                        {text('Dropdown Text', 'Dropdown')}
```
zero0Halo: I would actually not have this text configurable. It makes it seem as if it's a prop for `DropDown` but it isn't. I questioned why `DropDown` would have configurable text based on what I knew about how it was put together, I thought maybe you changed something. If it confused me, I think it'll confuse the average user too.

src/utils/position/position.ts
```diff
@@ -0,0 +1,105 @@
...
+            });
+        case 'topEnd':
+        default:
+            return css({
+                right: 0,
+                top: 0,
+                transform: `translate3d(0, -${overlayHeight + spacing}px, 0)`,
+            });
+    }
+};
```
zero0Halo, line 105: Love that you moved this out of `Tooltip` as its own resource to draw from.

src/DropDown/DropDown.stories.tsx
brandonlilly: I agree, but I also want a knob for the text so that I can easily test the behavior for various sizes. Maybe I could relabel it for clarity?

src/Tooltip/Tooltip.component.tsx
brandonlilly: It was the fastest way to extract the position utility without having to redo imports from this file.

But ultimately I figured it was alright to leave it here and be able to import from this file to see what positions exist. I don't feel strongly about it however, just seems more convenient.

src/DropDown/DropDown.stories.tsx
zero0Halo: Ah, makes sense from a testing perspective. I think any way you can denote that (even a comment in the story) would be good.

src/Tooltip/Tooltip.component.tsx
zero0Halo: Since you're doing it more for convenience in the developing of the stories I'm not going to be picky about it. My concern was more about the ultimate export of two components, if `Position` was being exported twice that could be a problem. As that's not what's going on, I'm not worried.

zero0Halo (approved):
LGTM

PR #115, opened 2019-09-06
feat: add generate component command

brandonlilly:

---------

Adds a `yarn generate MyComponent` command to generate a new component from a template.

### DIFF
--- a/scripts/MyComponent/MyComponent.stories.tsx
+++ b/scripts/MyComponent/MyComponent.stories.tsx
@@ -1,0 +1,24 @@
+// REACT
+import * as React from 'react';
+// STORYBOOK
+import { storiesOf } from '@storybook/react';
+// VENDOR
+import styled, { ThemeProvider } from '@xstyled/styled-components';
+import { th } from '@xstyled/system';
+// ANCHOR
+import { RootTheme } from '../../src/theme';
+import { MyComponent } from './MyComponent.component';
+
+const StyledStory = styled('div')`
+    padding: 2rem;
+    color: text.body;
+    font-family: ${th('typography.fontFamily')};
+`;
+
+storiesOf('Components/MyComponent', module).add('Default', () => (
+    <ThemeProvider theme={RootTheme}>
+        <StyledStory>
+            <MyComponent />
+        </StyledStory>
+    </ThemeProvider>
+));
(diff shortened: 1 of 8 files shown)


scripts/generateComponent.js
```diff
@@ -0,0 +1,45 @@
...
+        // copy template component
+        fs.copySync(srcPath, destPath);
+
+        // replace text in files
+        replace.sync({
+            files,
+            from: /MyComponent/g,
+            to: componentName,
+        });
+        replace.sync({
```
zero0Halo: I'm fuzzy with what's going on with this instance of `replace.sync`. I commented it out and the script appeared to work the same.

scripts/MyComponent/MyComponent.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+    font-family: ${th('typography.fontFamily')};
+`;
+
+storiesOf('Components/MyComponent', module).add('Default', () => (
+    <ThemeProvider theme={RootTheme}>
+        <StyledStory>
+            <MyComponent />
+        </StyledStory>
+    </ThemeProvider>
+));
```
zero0Halo, line 24: I think it'd be good to include at least one knob so the dev can expand on it.

scripts/generateComponent.js
```diff
@@ -0,0 +1,45 @@
...
+
+const fs = require('fs-extra');
+const argv = require('yargs').argv;
+const replace = require('replace-in-file');
+const Renamer = require('renamer');
+const renamer = new Renamer();
+
+const componentName = argv._[0];
+
+if (componentName) {
```
zero0Halo: In addition to checking if an argument was passed for the component name, it might also be good to have a check to ensure that the component doesn't already exist. Otherwise this script will still generate files in the destination folder. Attaching a screenshot of the two components I made, then running the script again. 

image

zero0Halo (commented):
I think an entry in the contributors page would be awesome. This is a great tool to help people get started.

scripts/MyComponent/MyComponent.stories.tsx
brandonlilly: I don't really want to do this because I'd want the knob to be for a prop, and there is no prop I'd want by default.

zero0Halo (approved):

PR #117, opened 2019-09-11
feat(Pagination): create Pagination component

brandonlilly:

---------

Creates a Pagination component

### DIFF
--- a/src/Pagination/Goto/Goto.components.tsx
+++ b/src/Pagination/Goto/Goto.components.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import classNames from 'classnames';
 import styled from '@xstyled/styled-components';
 // ANCHOR
-import { Input } from '../..';
+import { Input } from '../../Form';
 
 // Todo: this component needs work for mobile and for accessibility.
 
@@ -11,10 +11,17 @@ const EventKeyCodes = {
     ENTER: 13,
 };
 
-interface GotoProps {
+// This is a workaround to overwrite instead of extend, to avoid
+// an incompatible extension (for onSubmit)
+type Spread<L, R> = R & Pick<L, Exclude<keyof L, keyof R>>;
+
+interface ExtraProps {
     className?: string;
     onSubmit: (arg: number) => any;
 }
+
+interface GotoProps
+    extends Spread<React.HTMLAttributes<HTMLDivElement>, ExtraProps> {}
 
 const StyledGoto = styled('div')`
     width: 6rem;
--- a/src/Pagination/Goto/Goto.components.tsx
+++ b/src/Pagination/Goto/Goto.components.tsx
@@ -1,0 +1,63 @@
+// VENDOR
+import * as React from 'react';
+import classNames from 'classnames';
+import styled from '@xstyled/styled-components';
+// ANCHOR
+import { Input } from '../..';
+
+// Todo: this component needs work for mobile and for accessibility.
+
+const EventKeyCodes = {
+    ENTER: 13,
+};
+
+interface GotoProps {
+    className?: string;
+    onSubmit: (arg: number) => any;
+}
+
+const StyledGoto = styled('div')`
+    width: 6rem;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    padding: 0 0.5rem;
+
+    .anchor-input {
+        margin-left: 0.5rem;
+    }
+`;
+
+export const Goto = ({
+    onSubmit,
+    className,
+    ...props
+}: GotoProps): React.ReactElement<GotoProps> => {
+    const [value, setValue] = React.useState<null | number>(null);
+
+    const handleSubmit = () => {
+        if (onSubmit && value !== null) {
+            onSubmit(value);
+            setValue(null);
+        }
+    };
+
+    return (
+        <StyledGoto
+            className={classNames('anchor-pagination-goto', className)}
+            {...props}
+        >
+            Goto
+            <Input
+                type="number"
+                onChange={num => setValue(num)}
+                onKeyDown={event => {
+                    if (event.keyCode === EventKeyCodes.ENTER) {
+                        handleSubmit();
+                    }
+                }}
+                value={value === null ? '' : value}
+            />
+        </StyledGoto>
+    );
+};
--- a/src/Pagination/Pagination.component.tsx
+++ b/src/Pagination/Pagination.component.tsx
@@ -1,0 +1,217 @@
+// VENDOR
+import * as React from 'react';
+import classNames from 'classnames';
+import styled from '@xstyled/styled-components';
+// ANCHOR
+// import { colors } from '../theme';
+import { ButtonProps } from '../Button/Button.component';
+import { Goto } from './Goto';
+import {
+    Button as AnchorButton,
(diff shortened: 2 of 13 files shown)


src/Pagination/Goto/Goto.components.tsx
```diff
@@ -0,0 +1,63 @@
+// VENDOR
+import * as React from 'react';
+import classNames from 'classnames';
+import styled from '@xstyled/styled-components';
+// ANCHOR
+import { Input } from '../..';
```
sharkparty: Prolly wanna import this more declaratively.

src/Pagination/Goto/Goto.components.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+// Todo: this component needs work for mobile and for accessibility.
+
+const EventKeyCodes = {
+    ENTER: 13,
+};
+
+interface GotoProps {
```
sharkparty: Does this need the HTML attrs from React?

src/Pagination/Goto/Goto.components.tsx
```diff
@@ -0,0 +1,63 @@
...
+        margin-left: 0.5rem;
+    }
+`;
+
+export const Goto = ({
+    onSubmit,
+    className,
+    ...props
+}: GotoProps): React.ReactElement<GotoProps> => {
+    const [value, setValue] = React.useState<null | number>(null);
```
sharkparty, line 43: Why not always zero (`0`) this out? Is it because you'd like an active state in case the DOM displayed numbers do not include the active page, or some form of allowing a state without any active element?

src/Pagination/Pagination.component.tsx
```diff
@@ -0,0 +1,268 @@
...
+        dispatch({ type: 'totalChanged', total });
+    }, [total]);
+
+    React.useEffect(() => {
+        if (onChange) {
+            onChange(state.current);
+        }
+    }, [state.current]);
+
+    const pageButton = ({ page, slot }: { page: number; slot?: number }) => (
```
zero0Halo, line 149: It feels a little odd (maybe even anti-pattern?) to define a component within another component's render even if it is just a wrapper. Could there also be performance hits with re-renders?

src/Pagination/Pagination.component.tsx
```diff
@@ -0,0 +1,268 @@
...
+import { Goto } from './Goto';
+import {
+    Button as AnchorButton,
+    ChevronLeft,
+    ChevronRight,
+    Ellipses,
+    Typography,
+} from '..';
+
+// Constrain a value between a min and max
```
zero0Halo: It'd be beneficial to explain why. I feel like the code already demonstrates what is going on, but not why it's necessary.

src/Pagination/Goto/Goto.components.tsx
brandonlilly: Not sure exactly what you're asking, but `null | number` is effectively the Type of the `Input type="number"` value. I decided to keep that for the Type of the state here, with `null` representing an empty input.

src/Pagination/Pagination.component.tsx
```diff
@@ -0,0 +1,268 @@
...
+    pageSize = 10,
+    current: controlledCurrent,
+    showGoto,
+    showArrows = true,
+    size = 'sm',
+    onChange,
+    variant = 'pager',
+    ...props
+}: PaginationProps): React.ReactElement<PaginationProps> => {
+    const [state, dispatch] = React.useReducer(reducer, {
```
zero0Halo, line 126: This is cool, I haven't seen this hook yet.

src/Pagination/Goto/Goto.components.tsx
brandonlilly: Hmm
```
Interface 'GotoProps' incorrectly extends interface 'HTMLAttributes<HTMLDivElement>'.
  Types of property 'onSubmit' are incompatible.
    Type '(arg: number) => any' is not assignable to type '(event: FormEvent<HTMLDivElement>) => void'.
      Types of parameters 'arg' and 'event' are incompatible.
        Type 'FormEvent<HTMLDivElement>' is not assignable to type 'number'.
```

src/Pagination/Pagination.component.tsx
```diff
@@ -0,0 +1,268 @@
...
+    variant = 'minimal',
+    size = 'sm',
+    outline = false,
+    ...props
+}: ButtonProps) => (
+    <AnchorButton variant={variant} size={size} outline={outline} {...props} />
+);
+
+type Size = 'sm' | 'xs';
+type Variant = 'pager' | 'minimal';
```
zero0Halo: This is a very small quibble, but I feel that `pager` may not be the best wording as technically they're both still pagers. Maybe `default` and `minimal`?

brandonlilly: `default` is actually what I originally called it, but then I changed it because, what if we no longer want it to be the default option? I agree that `pager` isn't great, but honestly I have no idea what else to call it.

zero0Halo: That's fair. My other thought was `buttons`, since one view is using buttons and the other is... not. ;)

src/Pagination/Pagination.component.tsx
brandonlilly: Uh, I'm not really defining a component, just a closure. It's not the same because I'm actually calling the function, not returning it as a child element.

zero0Halo: Yeah, you're right, it is really a closure so that's my bad. I will admit I may be overly cautious on this. I've read a few articles where this approach was said to be poor for performance, but it's not like I've done serious perf testing on it.

zero0Halo (approved):

src/Pagination/Goto/Goto.components.tsx
sharkparty: I'm on the fence, between two tensions: (a) this should always have a value even if that value is 0 because you're always on a page in a paginated set or (b) allow the user to clear the value if they want.

brandonlilly: In the current route I went with, the Goto is empty and doesn't track the current page. If a user deigns to enter a value, then it'll go to that page and clear itself. (If the value the user entered it too small or large, it goes to the first or last page respectively. A value of `0` would be outside the range of viable pages to go to (and so you'd end up on the first page)

sharkparty: That behavior does make sense and we don't exactly have UX insight on this, so I'll resolve this. Thanks for the context!

sharkparty (approved):
LGTM

src/Pagination/Pagination.component.tsx
```diff
@@ -0,0 +1,270 @@
...
+        case 'totalChanged':
+            // This will make it so that if the current page is now beyond the
+            // new page total then the current page changes to the new last page.
+            // It might be that we want to revert to the first page instead.
+            return {
+                ...state,
+                current: constrain(1, state.current, action.total),
+            };
+        default:
+            throw new Error();
```
sharkparty: No-op: You could throw a message in here.

PR #121, opened 2019-09-20
feat(Alert): Alert Component

sharkparty:

---------

My code implements the alert component into Anchor.


src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+    warning = 'WARNING',
+    error = 'ERROR',
+}
+
+type AlertType =
+    | 'success'
+    | 'info'
+    | 'warning'
+    | 'error'
+    | AlertTypes.success
```
brandonlilly, line 31: Is there a particular raisin for doing both like this?

src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+    | AlertTypes.info
+    | AlertTypes.warning
+    | AlertTypes.error;
+
+interface AlertProps extends React.HTMLAttributes<HTMLDivElement> {
+    type: AlertType;
```
brandonlilly: What do you think of renaming `type` prop to `variant`? It's a bit more in line with other components, and gets rid of my nervousness about `type` possibly being a reserved word in some sense.

src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+            {...props}
+        >
+            {closable && (
+                <Button
+                    className="anchor-alert-close"
+                    size="xs"
+                    circular={true}
+                    variant="minimal"
+                    prefix={<Close color="rgba(0, 0, 0, 0.3)" scale="sm" />}
+                    onClick={() => handleClose()}
```
brandonlilly:
```jsx
onClick={handleClose}
```
?

src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+                />
+            )}
+            {icon ? (
+                <StyledIconContainer>
+                    {React.cloneElement(icon)}
```
brandonlilly: You could clone on a className to use that instead of the `.anchor-icon` selector, so that it works for more than just our icons. That or you could directly clone on the `color` prop instead of using css overrides. The `cloneWithProps` util might be of use.

src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+    return css`
+        ${({ theme }: any) =>
+            css({
+                backgroundColor: transparentize(
+                    0.75,
+                    get(theme, `colors.${type.toLowerCase()}`)
+                ),
+                border: `solid thin ${transparentize(
+                    0.4,
+                    get(theme, `colors.${type.toLowerCase()}`)
```
brandonlilly: There are semantic border colors for `error` and `success`. We could add the rest (`info`, `warning`, actually define them (currently set to placeholder of `ash.dark`) and use those.

brandonlilly: We could do the same for `colors.backgrounds` as well, if you like (e.g. `colors.backgrounds.error`.)

src/Alert/Alert.component.tsx
sharkparty: > raisin

lol

I want to support TS enums as well as plains JS strings.

src/Alert/Alert.component.tsx
sharkparty: ¯\_(ツ)_/¯

Fine by me. I was following what I saw in the wild, but this works. Will update now.

src/Alert/Alert.component.tsx
sharkparty: There's some complexity that arises because we'll need the same color but different levels of transparency in this component. Where would you put the transparency for the bg/border values? Or, would you reduce the complexity of those to use hex colors that render roughly the same hue?

src/Alert/Alert.component.tsx
sharkparty: Solid point here. I think the class-based approach would be the most flexible.

src/Alert/Alert.component.tsx
sharkparty: I just pushed a theme where I add the transparent colors to the theme to prevent introducing new hues to the color palette.

src/Alert/Alert.component.tsx
```diff
@@ -0,0 +1,178 @@
...
+                    circular={true}
+                    variant="minimal"
+                    prefix={<Close color="rgba(0, 0, 0, 0.3)" scale="sm" />}
+                    onClick={handleClose}
+                />
+            )}
+            {icon ? (
+                <StyledIconContainer>
+                    {React.cloneElement(icon, {
+                        className: 'anchor-alert-icon',
```
brandonlilly, line 168: This is pretty edge case, but this would override the className if a user passed one in. It's one reason I made the `cloneWithProps` util. e.g.
```jsx
icon={<InfoOutline scale="md" className="qa-alert-icon" />}
```
`qa-alert-icon` class would be lost

brandonlilly (approved):

src/Alert/Alert.component.tsx
sharkparty: Ah, that does make more sense. Pushing a fix now. Not sure why our CI is crashing. I was having a git issue earlier and I'm wondering if that corrupted something..

PR #124, opened 2019-09-24
docs(Grid): added to doc site

zero0Halo:
feat(ApiTable): passed data is sorted alphabetically by property value

feat(ApiTable): default column can accept a component instead of a string

docs: autocomplete, avatar, badge, button, card, collapse, hero & icon now all use ApiTable

feat(ApiTable): type column now can run parse() on string data

docs(input): converted table to ApiTable

docs(list): converted table to ApiTable

fix(Page): removed all css for styling the api tables

docs(ApiTable): updated documentation to note it now sorts alphabetically and removed the line about using semantic tags

docs(menu): using ApiTable

docs(modal): using ApiTable

docs(typography): using ApiTable

feat(ComponentInfo): outputs links to components on github

fix(ComponentInfo): better check for getting component data

feat(constants): new entry, COMPONENTS

fix(sections): changed Autocomplete to AutoComplete

chore(SiteLink): cleanup

chore(Page): cleanup

feat(ApiTable): sort can be disabled

chore(sections): Grid no longer hidden

docs(Wave): added component description

fix(AutoComplete): passed correct title to ComponentInfo

docs(grid): added api tables for Col, Container, Hidden, Row, ScreenBadge, ScreenClass and Visible

docs(grid): added code example

fix(build): gatsby has errors for exporting interfaces from index files, directly importing interfaces from sections

chore(SideNavigation): removed componentWillMount lifecycle method, used constructor instead as the former is being deprecated

chore(TemplateExample): removed console.log

docs(sections): Added proper description for grid

docs(grid): Changed approach, broke up components into 2 categories, Layout and Breakpoints

chore(typings): added entry for Container

fix(Page): changed breakpoints so that at medium the page still has the nav next to content

docs(grid): added Breakpoints section

docs(grid): added details on Breakpoint components

chore: prettier

---------

Adds documentation on the components belonging to `Grid`, converted all component docs to use the `ApiTable`, added some enhancements to `ApiTable`, changed breakpoints so that the non-mobile layout doesn't kick until under `md`.


docs/src/components/Utils/ApiTable.component.tsx
```diff
@@ -113,45 +116,75 @@ interface RowData {
...
-                                {Array.isArray(rowData.description)
-                                    ? rowData.description.map(
-                                          (description, i) => (
-                                              <React.Fragment key={`key-${i}`}>
-                                                  {typeof description ===
-                                                  'string'
-                                                      ? parse(description)
-                                                      : description}
-                                              </React.Fragment>
+const sortFn = (a: RowData, b: RowData) => {
```
sharkparty: I feel like this function could be more descriptively named, or—since this is basically a simple sort method—if could just be an inline method inside of the view-partial.

docs/src/components/Utils/ComponentInfo.component.tsx
```diff
@@ -2,14 +2,28 @@
...
 import { Typography } from '@retailmenot/anchor';
 // COMPONENTS
-import { sections, SiteLink } from '../Utils';
+import { COMPONENTS, sections, SiteLink } from '../Utils';
 
 const StyledArticle = styled('article')`
     padding: 1rem;
 `;
 
+const TitleLink = styled('a')`
```
sharkparty, line 14: Some of this could be props added to the `Typography` component. Why not add it that way?

zero0Halo: Originally  I did that, but changing the `scale` prop on `Typography` along with `a` as the `tag` doesn't change its size or line-height, so I went this route. You pointing it out here reminds me that I wanted to make a ticket to look into this as I wasn't sure if it was a bug or working-as-intended like the header tags.

sharkparty (approved):
LGTM

PR #125, opened 2019-09-26
feat(Menu): Updated Menu

sharkparty:

---------

My code updates the Menu and Item components and our library's testing suite.


src/DropDown/DropDown.component.tsx
```diff
@@ -170,16 +171,15 @@ export class DropDown extends React.Component<DropDownProps> {
     > = React.createRef();
     private rootElement: any;
     // Observable Subscriptions
-    private hoveredSub: Subscription;
-    private clickedSub: Subscription;
+    private hoveredSub$: Subscription;
+    private clickedSub$: Subscription;
```
zero0Halo, line 175: Why the addition of the `$` for these two? Consistency with the other variables?

src/Menu/Menu.component.tsx
```diff
@@ -3,119 +3,147 @@ import * as React from 'react';
...
+import { space as spaceStyles, SpaceProps } from '@xstyled/system';
 
-interface MenuProps extends React.HTMLAttributes<HTMLDivElement> {
+enum MenuSizes {
+    sm = 'sm',
+    md = 'md',
+    lg = 'lg',
+}
+
+export type MenuSizeProps = MenuSizes | 'sm' | 'md' | 'lg';
```
zero0Halo, line 18: Recently discovered that you can do something along the lines of 

```
const menuSizes  = {
    sm = 'sm',
    md = 'md',
    lg = 'lg',
}

export type MenuSizes = keyof typeof menuSizes;
```

Which will then give you the same effect as `interface MenuSizes = 'sm' | 'md' | 'lg';`

src/Menu/Menu.stories.tsx
```diff
@@ -7,11 +7,82 @@ import styled, { ThemeProvider } from '@xstyled/styled-components';
...
+            </ListItem>
+        ))}
+    </List>
+);
+
+const renderDemoMenu = () => {
+    const [isItemHovered, setIsItemHovered] = React.useState<boolean>(false);
+    const [isItem2Hovered, setIsItem2Hovered] = React.useState<boolean>(false);
+    return (
+        <>
```
zero0Halo: Do you really need the fragment since `Menu` is the only child?

zero0Halo (commented):
Please update the `DropDown` docs with the new `onTriggered` prop.

src/DropDown/DropDown.component.tsx
sharkparty: [sauce](https://cycle.js.org/basic-examples.html#basic-examples-increment-a-counter-what-is-the-convention)

src/Menu/Menu.component.tsx
sharkparty: I'd rather use the Typescript tool for solving this while also allowing pure JS persons to use strings than try to use an approach that doesn't give the advantages of either, but I have been doing some digging and by using the Enum it may be the case that the transpiled output isn't unlik what you're positing above, but instead of requiring an object's key, it'll accept the value of the keys in the JS version which would be a string. I'd need to build out a candidate to test this, but it's what I'm hoping for. That way TS gets an Enum and JS gets a string without any configuration.

src/Menu/Menu.stories.tsx
sharkparty: I do not. Good eye.

src/DropDown/DropDown.component.tsx
```diff
@@ -207,15 +207,18 @@ export class DropDown extends React.Component<DropDownProps> {
...
         const mouseLeave$ = fromEvent(dropDown, 'mouseleave');
 
-        this.clickedSub = merge(
+        this.clickedSub$ = merge(
             outsideClick$.pipe(mapTo(false)),
             escapeKey$.pipe(mapTo(false)),
             dropDownClick$.pipe(map(() => !this.state.clicked))
         ).subscribe(clicked => {
             this.setState({ clicked });
+            if (this.props.onTriggered) {
```
brandonlilly, line 216: I think this may fire unnecessarily, at times. e.g., the dropdown is closed, but the user hits the `esc` key.

In fact, this method would have issues when using the `both` or `hover` triggers, as it doesn't account for the `trigger` prop or how the two values combine the for `both` (e.g. you click to keep it opened, but hover and unhover).

I'd probably recommend `useEffect` but this isn't a functional component :/

src/Menu/Menu.stories.tsx
```diff
@@ -7,11 +7,80 @@ import styled, { ThemeProvider } from '@xstyled/styled-components';
...
+            <DropDown
+                overlay={<MockList />}
+                position="bottomStart"
+                showArrow={false}
+                trigger="click"
+                onTriggered={setIsItemHovered}
+            >
+                <Item>
+                    DropDown Link
+                    {isItemHovered ? (
```
brandonlilly, line 57: Def doesn't have to be this PR, but I kind of feel like we should have a component for this animated Chevron turn down/up thing. (Relevant here and to the Collapse).

src/Menu/__snapshots__/Menu.component.spec.tsx.snap
```diff
@@ -1,15 +1,67 @@
...
+}
+
+.c0 .anchor-menu-item:hover {
+  color: neutrals.white.base;
+  background-color: rgba(255,255,255,0.1);
+}
+
+.c0 .anchor-menu-item:active,
+.c0 .anchor-menu-item.active {
+  color: color:neutrals.white.base;
```
brandonlilly: Somethings up here

src/Menu/Menu.component.spec.tsx
```diff
@@ -4,31 +4,23 @@ import * as renderer from 'react-test-renderer';
...
-        console.log(wrapper.debug());
-        console.log(component.debug());
-
-        /* tslint:enable */
-        expect(subject).toBeDefined();
-        expect(wrapper).toBeDefined();
-        expect(component).toBeDefined();
+    it('should match its snapshot and render items.', () => {
+        const subject = (
+            <div>
```
brandonlilly: `ThemeProvider`?

src/Menu/Menu.stories.tsx
```diff
@@ -7,11 +7,80 @@ import styled, { ThemeProvider } from '@xstyled/styled-components';
...
+    const [isItemHovered, setIsItemHovered] = React.useState<boolean>(false);
+    const [isItem2Hovered, setIsItem2Hovered] = React.useState<boolean>(false);
+    return (
+        <Menu>
+            <DropDown
+                overlay={<MockList />}
+                position="bottomStart"
+                showArrow={false}
+                trigger="click"
+                onTriggered={setIsItemHovered}
```
brandonlilly, line 53: We could also solve this with [render props](https://reactjs.org/docs/render-props.html). Not sure if we actually should, but food for thought.

src/Menu/Menu.component.tsx
zero0Halo: I definitely can appreciate it using the value of the key rather than the key itself. I've been trying to find an approach that lets us not have to define the values for `type`'s twice, and also lets us export those values easily for use, such as for knobs.

src/DropDown/DropDown.component.tsx
sharkparty: I think this can be an optimization. I did notice that in those cases the event fires and the caret isn't predictable, but the way MegaMenu is presently wired this will work.

brandonlilly: But the second point- the behavior is broken for `both` and `hover`. Unless you're saying we can hold off on fixing it until we need it for more than just `click`

src/Menu/Menu.component.spec.tsx
sharkparty: 👍

src/Menu/__snapshots__/Menu.component.spec.tsx.snap
brandonlilly: This seems to still be present

src/Menu/Menu.component.tsx
```diff
@@ -3,119 +3,147 @@ import * as React from 'react';
...
+                })};
+            background-color: rgba(255, 255, 255, 0.1);
+        }
 
-    &:first-of-type {
-      margin-left: 0;
-    }
+        &:active,
+        &.active {
+            color: ${({ color }) =>
```
brandonlilly, line 100: The initial `color: ` is redundant.

sharkparty: 👍

src/DropDown/DropDown.component.tsx
sharkparty: The latter.

src/Menu/Menu.component.tsx
```diff
@@ -3,119 +3,147 @@ import * as React from 'react';
...
-
-  /* TODO: use React.clone to append a run-time class here instead of using a * selector */
-  > * > a,
-  > a {
-    display: inline-block;
-    line-height: 1.125rem;
-    ${({ color = 'neutrals.white.base' }) =>
+    display: flex;
+    width: 100%;
+    ${spaceStyles};
```
brandonlilly, line 71: This is probably my bad, but 
```
${spaceStyles};
```
should go after
```
margin: 0;
padding: 0
```

so that it doesn't get overritten

src/Menu/Menu.component.tsx
brandonlilly: I think you marked this resolved, but it hasn't been updated in `origin` yet

brandonlilly (approved):

PR #128, opened 2019-09-26
fix(Typography): fix scale precedence, dedup redundant styles, remove margins, hue prop

brandonlilly:

---------

Makes `scale` take precedence over `tag` (so that they can be used usefully together)
Removes most redundant styles
Removes typography margins
Removes hue prop

### DIFF
--- a/src/Typography/Typography.component.tsx
+++ b/src/Typography/Typography.component.tsx
@@ -3,14 +3,18 @@ import * as React from 'react';
 // VENDOR
 import classNames from 'classnames';
 import styled, { css } from '@xstyled/styled-components';
-import { th, variant } from '@xstyled/system';
+import {
+    //
+    th,
+    variant,
+} from '@xstyled/system';
 // ANCHOR
 import { space as spaceStyles, SpaceProps } from '@xstyled/system';
 // import { colors, Color } from '../theme';
 import { TypographyTags, Scale } from '../theme/typography.theme';
 import { rem } from '../utils/rem/rem';
 
-type FontWeights =
+export type FontWeights =
     | 'normal'
     | 'bold'
     | 'bolder'
@@ -27,7 +31,7 @@ type FontWeights =
     | 'initial'
     | 'inherit';
 
-type TextTransforms =
+export type TextTransforms =
     | 'none'
     | 'capitalize'
     | 'uppercase'
@@ -60,7 +64,14 @@ type DisplayValues =
     | 'initial'
     | 'inherit';
 
-export type TextAlign = 'center' | 'left' | 'right' | 'inherit' | 'justify';
+export type TextAlign =
+    | 'center'
+    | 'left'
+    | 'right'
+    | 'inherit'
+    | 'justify'
+    | 'start'
+    | 'end';
 
 export interface TypographyProps extends SpaceProps, React.HTMLAttributes<any> {
     className?: string;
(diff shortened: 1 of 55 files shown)


src/Typography/Typography.component.tsx
```diff
@@ -3,14 +3,18 @@ import * as React from 'react';
 // VENDOR
 import classNames from 'classnames';
 import styled, { css } from '@xstyled/styled-components';
-import { th, variant } from '@xstyled/system';
+import {
+    //
```
zero0Halo: Guessing a leftover artifact?

src/Typography/Typography.component.tsx
```diff
@@ -3,14 +3,18 @@ import * as React from 'react';
...
+    th,
+    variant,
+} from '@xstyled/system';
 // ANCHOR
 import { space as spaceStyles, SpaceProps } from '@xstyled/system';
-import { colors, Color } from '../theme';
+// import { colors, Color } from '../theme';
```
zero0Halo: Any reason why this is still here?

src/Typography/Typography.stories.tsx
```diff
@@ -2,36 +2,75 @@
...
+    'inherit',
 ];
+const transforms = [
+    'none',
+    'capitalize',
+    'uppercase',
+    'lowercase',
+    'initial',
+    'inherit',
+] as TextTransforms[];
```
zero0Halo, line 59: `alignments`, `weights` and `transforms` probably could all benefit from the `keyof typeof` technique we talked about earlier this week so that you don't have to  duplicate the info that's already in the types in `Typography`.

src/Typography/Typography.stories.tsx
```diff
@@ -64,4 +106,35 @@ storiesOf('Components/Typography', module)
                 </div>
             </StyledStory>
         </ThemeProvider>
-    ));
+    ))
+    .add('Knobus Propus', () => {
```
zero0Halo, line 110: Nice.

src/Typography/Typography.stories.tsx
```diff
@@ -64,4 +106,35 @@ storiesOf('Components/Typography', module)
...
+        const content = text(
+            'text',
+            'The quick brown fox jumps over the lazy dog.'
```
zero0Halo, line 113: The quick red fox jumps over the brown lazy dog. That sentence uses every letter in the American alphabet.

src/Typography/Typography.component.tsx
```diff
@@ -72,46 +83,45 @@ export interface TypographyProps extends SpaceProps, React.HTMLAttributes<any> {
...
+        })}
 
-    // CSS Overrides
-    ${({ align = 'inherit', display, transform = 'none' }) =>
+    // Prop overrides. We don't have any defaults here because then they
+    // would always override the variants above.
+    ${({ color, align, display, transform, lineHeight, size, weight }) =>
         css({
+            color,
             textAlign: align,
```
sharkparty, line 117: No-op, but I probably should have left `align` as `textAlign`.

src/Typography/Typography.stories.tsx
brandonlilly: The `red` is unnecessary. Speaking of red, did you know that strawberr......

src/Typography/Typography.component.tsx
brandonlilly: I was thinking along the same lines. Same goes for `weight` `fontWeight`, `transform` `textTransform`, and `size` `fontSize`.

We could allow both for each, and optionally deprecate.

src/Typography/Typography.component.tsx
brandonlilly: woops

src/Typography/Typography.component.tsx
sharkparty: I'll leave that to your discretion. In my mind that makes me wonder (as an engineer) why not just let me alter the CSS via the `style` attribute instead..

brandonlilly: I'll leave it for now.

brandonlilly: I was also considering allowing a string for the `size` and `lineHeight` props, to allow for more than just `rem` values.

src/Typography/Typography.stories.tsx
brandonlilly: I think I'm gonna skip that for now while I don't fully understand the implications.

zero0Halo (approved):

PR #132, opened 2019-10-01
fix(Button): only show focus state on tab

brandonlilly:

---------

Makes the outline only appear when the button is tabbed to.

### DIFF
--- a/src/Button/Button.component.tsx
+++ b/src/Button/Button.component.tsx
@@ -36,7 +36,6 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
     block?: boolean;
     circular?: boolean;
     reverse?: boolean;
-    outline?: boolean;
 
     prefix?: any;
     suffix?: any;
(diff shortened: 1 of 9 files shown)


src/Button/Button.component.tsx
```diff
@@ -502,8 +482,13 @@ export const Button = ({
     minWidth,
     prefix,
     suffix,
+    onMouseDown,
+    onMouseUp,
```
sharkparty, line 486: This probably isn't an issue, but would we want onTouchStart/End to mirror this behavior?

src/Button/Button.component.tsx
```diff
@@ -567,10 +552,33 @@ export const Button = ({
...
+                    onMouseUp(event);
+                }
+            }}
+            onFocus={event => {
+                if (mouseDown) {
+                    // This keeps the button from being :focused when
+                    // clicked so that it is only applied when tabbed to.
+                    // We want the outline to appear when tabbing for
+                    // accessibility.
+                    event.target.blur();
```
sharkparty, line 573: Good note. The above methods could be consolidated into a single handler with a binary flag.

sharkparty: Also, TS may flip out over the absence of type declaration on the `event` arg. Surprised your linter didn't scream bloody murder.

sharkparty (approved):
LGTM

PR #134, opened 2019-10-02
docs: Update Card and DropDown to use Anchor components in examples

zero0Halo:
chore(typings): added Card, InfoOutline, List, ListDivider & ListTitle

fix(MoreInfo): using DropDown for its example rather than homebrew

fix(MyList) using List for its example rather than homebrew

docs(card and dropdown): Updated code blocks based on the above changes

chore: prettier

---------

My code does...

### DIFF
--- a/docs/src/typings.d.ts
+++ b/docs/src/typings.d.ts
@@ -11,6 +11,7 @@ declare module '@retailmenot/anchor' {
   const AutoComplete: any;
   const Badge: any;
   const Button: any;
+  const Card: any;
   const Cart: any;
   const Check: any;
   const Col: any;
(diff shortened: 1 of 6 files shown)


docs/src/typings.d.ts
```diff
@@ -11,6 +11,7 @@ declare module '@retailmenot/anchor' {
   const AutoComplete: any;
   const Badge: any;
   const Button: any;
+  const Card: any;
```
sharkparty: We actually ship anchor with its own typings file so this shouldn't be needed so long as the client is on the latest version of Anchor. Also, typing anything `any` isn't a great idea. It's good for accelerating against a tough deadline, but it's not a sustainable practice.

sharkparty (changes requested):
Let's try and figure out the typings issue.

docs/src/typings.d.ts
zero0Halo: I'll look into this. If we're shipping the types, it's really odd they aren't coming through. Agreed about `any`, this list has just been building and building. Need to kill the trend.

zero0Halo: For posterity, @sharkparty discovered that removing the Anchor type definitions in the repo solves the issue. Most likely these were given priority over the definitions in the Anchor module.

sharkparty (approved):
LGTM

PR #135, opened 2019-10-04
chore: extract colors

brandonlilly:

---------

My code does...


.storybook/addons.js
```diff
@@ -1,6 +1,6 @@
 /* tslint:disable */
+import '@storybook/addon-knobs/register';
```
brandonlilly, line 2: This makes Knobs the first addon tab

src/Alert/__snapshots__/Alert.spec.tsx.snap
```diff
@@ -53,7 +53,7 @@ Array [
   font-weight: normal;
   text-align: inherit;
   color: inherit;
-  color: #595959;
```
brandonlilly: `#595959` (charcoal.light) comes up as a text color often enough that I think it should be added as a `text.???` theme entry, but I'm having trouble deciding what to call it.

src/Icon/Icons/AddEvent.component.tsx
```diff
@@ -19,6 +19,7 @@ export const AddEvent = ({
     <StyledIcon
         className={classNames('anchor-icon add-event', className)}
         scale={scale}
+        $color={color}
```
brandonlilly, line 22: I changed the way this works so that the `svg` is always `currentColor` and the icon color is controlled instead by the `color` of the surrounding span. This is so that it can go through the theme interpolator (`fill={color}` won't).

If this method is not preferred I could manually run it through the theme instead.

brandonlilly: `$color={color}` to avoid `color=` on the DOM

src/Pagination/Pagination.spec.tsx
```diff
@@ -51,32 +50,4 @@ describe('Component: Pagination', () => {
         const tree = renderer.create(subject).toJSON();
         expect(tree).toMatchSnapshot();
     });
-    it('should behave properly when uncontrolled.', () => {
```
brandonlilly, line 54: This test was too fragile, so I removed it.

src/Skeleton/Skeleton.component.tsx
```diff
@@ -19,6 +19,8 @@ interface SkeletonProps
     minHeight?: string;
     minWidth?: string;
     textLength?: number;
+    colorA?: string;
```
brandonlilly: Added `colorA` and `colorB` as props for the two colors that Skeleton oscillates between. Open to other suggestions for names (`colorStart`, `colorEnd`?)

src/Typography/Typography.component.tsx
```diff
@@ -84,7 +84,13 @@ export interface TypographyProps extends SpaceProps, React.HTMLAttributes<any> {
     lineHeight?: number | string;
 }
 
-const StyledTypography = (tag: TypographyTags) => styled(tag)<TypographyProps>`
+export interface StyledTypographyProps extends TypographyProps {
+    $color?: 'inherit' | 'initial' | string;
```
brandonlilly, line 88: Added `$color` here as well to keep it out of the dom

src/theme/colors.theme.ts
```diff
@@ -165,6 +166,7 @@ export const ColorsTheme: ColorsThemeSchema = {
     },
     text: {
         body: colors.charcoal.base,
+        meta: colors.ash.dark,
```
brandonlilly: Named based on the roux palette. Open to suggestions.

src/Alert/__snapshots__/Alert.spec.tsx.snap
sharkparty: [Bootstrap](https://getbootstrap.com/docs/4.3/utilities/colors/) has a few semantic names we can pull from. Is there a common usage pattern with that color that offers a potential name?

src/Icon/Icons/AddEvent.component.tsx
sharkparty: I like this. It's a clever optimization, though, so we may want to document the reason for this approach in little more than what you've already outlined in the readme/docs site

src/Skeleton/Skeleton.component.tsx
sharkparty: `colorStart` and `colorEnd` are more semantic, so I prefer those.

src/Typography/Typography.component.tsx
sharkparty: I really don't like the idea of embedding a special character into the props. Where did you derive this from?

brandonlilly: This is a pattern I used in `Button` component for `fontSize`, `height`, and `size`. I based it off of the discussion in this thread https://github.com/styled-components/styled-components/pull/2093 (which afaict, still doesn't have a definite resolution). It's merely internal to the component, so the consumer still just uses the `color` prop when using `Typography`.

src/Skeleton/Skeleton.component.tsx
brandonlilly: I agree. The only hesitation I had was that the colors actually oscillate- a fact which `start` and `end` don't really capture. I'll make the change.

sharkparty: Just because something's looped doesn't mean it lacks and start and end, especially in this case.

src/Typography/Typography.component.tsx
sharkparty: I feel like `fontColor` is a bad idea because it's substandard, but that this creates confusion in a code-case where observables may be present. Since there's no clear winner here, I defer to your judgement. While I think the `$` is likely to raise more questions than a substandard name like `fontColor`, `$` is less technically misleading as it doesn't create a intermediate standard between the DOM and the component props. ¯\\_(ツ)_/¯

sharkparty (approved):
LGTM

docs/src/components/Layout/Page/Page.component.tsx
```diff
@@ -154,11 +154,37 @@ const HamburgerCol = styled(CenteredCol)`
     max-width: 3rem;
 `;
 
-const AnchorTheme = {
-    ...RootTheme,
-    fonts: {
-        ...RootTheme.fonts,
-        mono: `'SFMono-Regular', Consolas, Liberation Mono, Menlo, Courier, monospace`,
+const AnchorDocsTheme = merge({}, RootTheme, {
```
zero0Halo, line 157: Wondering why we need a 3rd party lib for this. Can accomplish the same thing with spread syntax, although granted it's a little messier.

docs/src/pages/components/badge.mdx
```diff
@@ -219,7 +219,7 @@ export const BadgeWithParentExample = () => {
         property: 'textColor',
         description: <ColorBlurb defaultTheme label="count" />,
         type: <FormatTypes data={['string', 'rgb', 'rgba']} noInterpret />,
-        default: <Link to="/theme/theme-provider#colors">white.base</Link>,
+        default: <Link to="/theme/theme-provider#colors">white</Link>,
```
zero0Halo, line 222: Since this is now a string literal rather than an abstracted value, no reason to link to the theme provider.

docs/src/pages/theme/colors.mdx
```diff
@@ -1,30 +1,4 @@
-import { ColorSwatchFamilyList } from '../../components/ColorSwatchFamilyList';
-import { Link } from 'gatsby';
+# Colors
 
-# colors
+This Page is Coming Soon...
```
zero0Halo, line 3: I feel like perhaps we should consider keeping the content of this page but putting a large note that it is not applicable to Anchor v1.0.1 and up rather than killing it entirely.

src/Skeleton/Skeleton.stories.tsx
```diff
@@ -101,7 +101,12 @@ storiesOf('Components/Skeleton', module)
                         tag={tag || undefined}
                         scale={scale || undefined}
                     >
-                        <Skeleton loading={loading} textLength={textLength}>
+                        <Skeleton
+                            loading={loading}
+                            textLength={textLength}
+                            colorStart={text('colorStart', '') || undefined}
+                            colorEnd={text('colorEnd', '') || undefined}
```
zero0Halo, line 108: Use the `colors` knob rather than `text` if possible. Nicer user experience.

src/theme/index.ts
```diff
@@ -37,9 +24,14 @@ export const RootTheme = {
...
-        base: `thin solid ${colors.ash.light}`,
-        dark: `thin solid ${colors.ash.dark}`,
+        light: `thin solid ${ColorsTheme.borders.light}`,
+        base: `thin solid ${ColorsTheme.borders.base}`,
+        dark: `thin solid ${ColorsTheme.borders.dark}`,
+    },
+
+    skeleton: {
+        colorStart: '#E7E7E7',
+        colorEnd: '#D3D3D3',
```
zero0Halo, line 34: Is it necessary to put these values here? No other components have them inserted directly into the theme like this.

zero0Halo (commented):
In the doc site, most (all?) code examples are not correctly interpolating the color values. I'm guessing `ThemeProvider` and `AnchorDocsTheme` need to be added to the `CodePreview` component. I will look into it since @brandonlilly is out.

EDIT: I stand corrected. Ran an anchor build and dropped the updated modules into the doc site. All is well. Nice job, Brandon.

zero0Halo:
We'll need to bump the version to 1.0.3 then do a quick turn around updating the doc site with that version.

sharkparty:
This will actually be version `1.1.0` since it's a pretty large series of breaking changes along the theme provider.

PR #141, opened 2019-10-22
fix: doc site mobile view

zero0Halo:
chore: added react-awesome-style-grid directly to use `config` export

Should be removed as soon as Anchor's implementation of a new grid system is in place

fix(index, Page): used config instead of responsiveCss

chore(AddBreakpoints): deleted as helper function is no longer used

fix(Page): refactored logic to show/hide the section nav with css and props

chore(Utils): cleanup

fix(Page): added custom breakpoints to `awesomegrid` entry for theme

feat(rem): added new utility function that converts px to rem

fix(homepage): removed ScreenBadge

chore(typings): added config from react-awesome-styled-grid

chore: prettier

---------

The `ScreenClass`, `Visible` and `Hidden` components from `react-awesome-styled-grid` (RASG) use `ComponentWillMount`, which Gatsby does not fire during its initial render. The site looks broken on mobile on a page's first load because of this.

I ended up using the `config` export from RASG to generate media queries and updated the `AnchorDocsTheme` with breakpoints from Anchor's `StandardBreakpoints` export.


docs/src/components/Layout/Page/Page.component.tsx
```diff
@@ -96,35 +89,55 @@ const StyledHeader = styled('header')`
...
+    height: 120%;
+    left: 0;
+    overflow-y: scroll;
+    padding-bottom: 30rem;
+    position: fixed;
+    right: 0;
+    top: 4.5rem;
+    width: 100vw;
+    z-index: 90;
+    display: ${({ sectionNavOpen }) => (sectionNavOpen ? 'display' : 'none')}
```
sharkparty: Has this been linted? Curious about the indentation. Also, if we're setting the `display` prop value we should use the xstyled-comp sytax: `${(props) => css({ display: props ? 'block' : 'none' })}` or `${(props) => css'display: ${props ? 'block' : 'none'}'` instead of `display: ${(props) => props ? 'block' : 'none'`

sharkparty: pretend quotes are backticks.

zero0Halo: Yep, this is post-lint. What's the `css` module give us in this case? I don't have an issue with using it, but unclear as to the why

sharkparty: We aren't overriding a string, per se, we're overriding a css object. IIRC xstyled prefers this and it won't throw a warning in our TS compiler about type mismatches.

sharkparty (approved):
LGTM

PR #144, opened 2019-10-25
feat: adds search to docsite

zero0Halo:
feat(rgba): new util. takes a hex code and opacity level and returns a css rgba string

chore(SearchInput): converted many css elements into styled components

chore(gitignore): added docsearch-scraper directory

fix(Page): used named export for Helmet

feat(scrape): added scrape script to package.json

chore: cleanup

feat(Page): added accent colors to the theme

fix(SectionNav): used accent colors from theme

fix(Page): better responsive breakpoint implementation

chore(Page): search is disabled for mobile

Clearly this is not ideal, but want to roll search out then add this in

chore: added algoliasearch and and lodash.debounce modules

feat(SearchInput): complete rewrite. Now directly uses Algolia api to hit docsearch index

feat(Results): new component. used by SearchInput as a template for search results.

chore: code cleanup

---------

This adds search to the doc site. This works in two parts:

* https://anchor.retailmenot.design/ is added to the DocSearch registry and managed by Algolia. They routinely scrape our site and populate an Algolia index which (for now) only I have direct access to.
* Using Algolia's API, I query the index and feed that info into an `AutoComplete` component.

The biggest caveat is that I'm hiding the search box on mobile. That's a big TODO, but this push is the first step.

### DIFF
--- a/docs/src/components/Layout/Page/Page.component.tsx
+++ b/docs/src/components/Layout/Page/Page.component.tsx
@@ -26,6 +26,7 @@ import {
     EnhancedNextPrevious,
 } from '../../Navigation';
 import { MDXComponents } from './MDXComponents';
+import { SearchInput } from '../../SearchInput';
 import { breakpoints, rem } from '../../Utils';
 // ASSETS
 import logo from './anchor-logo.svg';
@@ -171,6 +172,20 @@ const HamburgerCol = styled(CenteredCol)`
     `}
 `;
 
+const LogoCol = styled(CenteredCol)`
+    padding-right: 0;
+`;
+
+// TODO: This is a stopgap measure to get search out. The current mobile layout, plus the way
+// DocSearch initializes, makes both the UX and UI for mobile search tricky. Hiding it for mobile
+// for now
+const SearchCol = styled(CenteredCol)`
+    display: none;
+    ${props => config(props).media[breakpoints.md]`
+        display: flex;
+    `}
+`;
+
 const AnchorDocsTheme = merge({}, RootTheme, {
     colors: {
         primary: {
@@ -182,6 +197,11 @@ const AnchorDocsTheme = merge({}, RootTheme, {
             base: '#EB663D',
             light: '#FA9F5A',
             dark: '#C74E30',
+        },
+        accent: {
+            base: '#E3EEF6',
+            light: '#0998D6',
+            dark: '#0074A6',
         },
         borders: {
             light: '#F1F1F1',
@@ -255,20 +275,20 @@ export const Page = ({
                                 />
                             </HamburgerCol>
 
-                            <CenteredCol md={2} xs={2} sm={4}>
+                            <LogoCol md={2} lg={3} xl={2}>
                                 <StyledLogoContainer to="/">
                                     <img
                                         alt="Anchor Logo Horizontal"
                                         src={logo}
                                     />
                                 </StyledLogoContainer>
-                            </CenteredCol>
+                            </LogoCol>
 
-                            <CenteredCol md={2} xs={1} sm={1}>
-                                {/* TODO: Search will go here */}
-                            </CenteredCol>
+                            <SearchCol>
+                                <SearchInput />
+                            </SearchCol>
 
-                            <CenteredCol md={4} xs={1} sm={3}>
+                            <CenteredCol>
                                 <StyledPrimaryNav>
                                     <Typography
                                         tag="a"
@@ -285,7 +305,7 @@ export const Page = ({
                 </StyledHeader>
 
                 <Row>
-                    <SectionNavCol md={2} lg={2}>
+                    <SectionNavCol md={2} lg={3} xl={2}>
                         <FixedBody sectionNavOpen={sectionNavOpen} />
 
                         <StyledSectionNav sectionNavOpen={sectionNavOpen}>
@@ -293,7 +313,7 @@ export const Page = ({
                         </StyledSectionNav>
                     </SectionNavCol>
 
-                    <Col md={6} lg={10}>
+                    <Col md={6} lg={9} xl={10}>
                         <StyledContentMain>
                             <br />
                             <MDXProvider components={MDXComponents}>
--- a/docs/src/components/SearchInput/SearchInput.component.tsx
+++ b/docs/src/components/SearchInput/SearchInput.component.tsx
@@ -1,0 +1,164 @@
+/*
+    This component leverages Algolia's search API along with their DocSearch service.
+    http://anchor.retailmenot.design is scraped every 24 hours by DocSearch, and that scraped data
+    is fed into an Algolia index which, for now, only I (steve.swanson@rmn.com), have direct access
+    to. This component uses Algolia's API to then query that index to provide the user with results.
+*/
+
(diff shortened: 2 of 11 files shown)


docs/src/components/Layout/Page/Page.component.tsx
```diff
@@ -255,20 +275,20 @@ export const Page = ({
...
                                     />
                                 </StyledLogoContainer>
-                            </CenteredCol>
+                            </LogoCol>
 
-                            <CenteredCol md={2} xs={1} sm={1}>
-                                {/* TODO: Search will go here */}
-                            </CenteredCol>
+                            <SearchCol>
+                                <SearchInput />
```
brandonlilly, line 288: When I select a result and go to the linked page, the Search Input doesn't seem to exist anymore..

brandonlilly: Ah, just realized thats because the links when I run it locally are pointing at the live prod site.

docs/src/components/SearchInput/SearchInput.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+                /*
+                    Don't want titles repeated if they're right next to each other. The order of
+                    Algolia's results appear to be based on relevance, so it's possible that further
+                    down the list the same title will appear.
+                */
+                if (hitSection !== previousSection) {
+                    dataSource.push({
+                        label: hitSection,
+                        id: `${hit.objectID}-title`,
```
brandonlilly, line 118: If you press the down arrow key to go tab through the results, the titles (e.g. `COMPONENTS`) are included in the autocomplete selections. If you add `listItemType: 'title'` here, it'll solve the issue.

docs/src/components/SearchInput/Results.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+    value?: Values;
+}
+
+export const Results = ({
+    index,
+    currentIndex,
+    label,
+    value,
+}: ResultsProps) => {
+    const hasLink = value ? value.url : false;
```
brandonlilly: You can then do this logic, not by whether it has a link but by the `listItemType`

docs/src/components/SearchInput/SearchInput.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    render() {
+        const { searchIndex } = this;
+        const { dataSource } = this.state;
+
+        return (
+            <AutoComplete
+                placeholder="Search"
+                prefix={<SearchIcon />}
```
brandonlilly:
```jsx
<SearchIcon color="borders.base"/>
```
? :)

docs/src/components/SearchInput/SearchInput.component.tsx
zero0Halo: That's interesting. I wonder if there's a problem with my Results template because it should take priority; it shouldn't be necessary to declare the `listItemType` in the dataSource being fed to it.

brandonlilly: Your `Results` template is being used, but the `Autocomplete` uses the `dataSource` to determine which items should be selectable/tabbable, so you still need to specify the `listItemType`

docs/src/components/SearchInput/SearchInput.component.tsx
zero0Halo: I tried this and `borders.dark` out, but there isn't enough contrast IMO for the icon to be all that noticeable.

docs/src/components/SearchInput/Results.component.tsx
zero0Halo: Just based on behavior and debugging on the doc site, it doesn't appear that `listItemType` gets passed to a custom result template. Bears investigating with the `AutoComplete` component. Until then I'm going to keep the current logic.

brandonlilly: Ah yeah, looks like an oversight to me. Should likely be a 1 liner to pass it in.

zero0Halo:
Realized I forgot a huge part of this user-experience-wise. Using the arrows to select a result then hitting enter should take them to that entry. I've added that in with the latest commit.

docs/src/components/SearchInput/SearchInput.component.tsx
```diff
@@ -119,18 +139,31 @@ export class SearchInput extends React.PureComponent<InitialState> {
         }
     }
 
+    // Using the label that was selected, uses the associated url and redirects to that page.
+    goTo(value: string) {
+        const result = this.state.dataSource.find((item: DataSet) => {
```
brandonlilly, line 139: I don't think we'll run into any issues with our current layout, but I can imagine this having potential collisions between results with the same name e.g.
```
OVERVIEW
Introduction
CONTRIBUTE
Introduction
```
Clicking the Contribute Introduction would link to the Overview Introduction url. Although, if you do this, I don't think there's any more need for a custom `resultTemplate` (since you have the click behavior defined now).

I think these Autocomplete event handlers (onSelect, etc) should be refactored to pass along the full item information. I also think the label should eventually allow more than just a `string`. But thats outside the scope of this PR.

docs/src/components/SearchInput/SearchInput.component.tsx
brandonlilly: I had tested it locally, and it is noticeable to me, but its just a nitpick

docs/src/components/SearchInput/SearchInput.component.tsx
zero0Halo: I agree. I was surprised that only the value of `label` was passed, it makes more sense for the payload be the object. I haven't dug into `AutoComplete` yet, but I'm wondering if only the value of the `input` is being passed (which makes sense in a way). For now, since it is using `find` it'll only ever return the first match so this isn't breaking. Worth making a ticket for `AutoComplete` because ultimately I think that's a good enhancement for it to have.

zero0Halo: I should add that the result template is still necessary because I think allowing the user to click on the result should still be an option. That behavior doesn't exist in `ListItem` otherwise.

brandonlilly: Click behavior still worked when I tried it (removing `resultTemplate={Results}`) locally.

zero0Halo: I stand corrected. I didn't realize the `onSelect` prop worked for both selection via hitting `enter` and also for clicking a result. I removed the custom results template.

docs/src/components/SearchInput/SearchInput.component.tsx
```diff
@@ -156,7 +156,7 @@ export class SearchInput extends React.PureComponent<InitialState> {
                 onFilter={searchIndex}
                 onSelect={goTo}
                 placeholder="Search"
-                prefix={<SearchIcon />}
+                prefix={<SearchIcon pl="2" />}
```
brandonlilly, line 159: That looks much better

brandonlilly (approved):

PR #146, opened 2019-10-28
fix: autocomplete input warning

zero0Halo:

---------

This fixes a very loud warning from React regarding uncontrolled components for the `Input` component, which in turn fixes the same warning from `AutoComplete` since it uses `Input`.

### DIFF
--- a/src/Form/Input/Input.stories.tsx
+++ b/src/Form/Input/Input.stories.tsx
@@ -16,11 +16,19 @@ const StyledStory = styled('div')`
     padding: 2rem 5rem;
 `;
 
+type InputState = {
+    text?: string;
+    number?: string | number;
+    tel?: string | number;
+};
+
 const InputStories = () => {
     const { useState } = React;
-    const [textInputVal, setTextInputVal] = useState<string>('');
-    const [numberInputVal, setNumberInputVal] = useState<number | string>('');
-    const [phoneInputVal, setPhoneInputVal] = useState<number | string>('');
+    const [inputValues, setInputValues] = useState<InputState>({
+        text: '',
+        number: '',
+        tel: '',
+    });
 
     return (
         <ThemeProvider theme={RootTheme}>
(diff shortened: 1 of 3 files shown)


src/Form/Input/Input.stories.tsx
```diff
@@ -18,7 +18,10 @@ const StyledStory = styled('div')`
 
 const InputStories = () => {
     const { useState } = React;
-    const [someInputVal, setSomeInputVal] = useState<string>('');
+    const [textInputVal, setTextInputVal] = useState<string>('');
```
brandonlilly: Whats the reasoning behind splitting this into 3 disparate states?

zero0Halo: That's a fair point. I'll consolidate them into a single state.

src/Form/Input/Input.stories.tsx
```diff
@@ -16,9 +16,20 @@ const StyledStory = styled('div')`
...
+type InputState = {
+    text?: string;
+    number?: string | number;
+    tel?: string | number;
+};
+
 const InputStories = () => {
     const { useState } = React;
-    const [someInputVal, setSomeInputVal] = useState<string>('');
+    const [inputValues, setInputValues] = useState<InputState>({
```
brandonlilly, line 27: This is probably better, but was I was actually asking about is why theres a need to have different state for each of them, as opposed to all just using the same state like before.

zero0Halo: I did that at first. If they all pull from a single state value then all the form fields update at the same time with the same value. Originally only the `text` input was using state, and typing into the other fields was giving an uncontrolled form field warning.

brandonlilly: Oh derp. This is the story, not the component file. My bad, nothing to see here.

brandonlilly (approved):

PR #150, opened 2019-11-06
feat(AutoComplete): Expose Input component

zero0Halo:
feat: adds a new prop to AutoComplete, inputComponent. Defaults to Input.

---------

This adds a new prop, `inputComponent` to `AutoComplete`. It defaults to use the `Input` component, but it allows the user to pass their own `Input` component provided they set up a `forwardRef`. Comments are in the code, documentation updates forthcoming.

This also fixes a rendering issue with Gatsby for `Grid`. Gatsby hates `window` and although I tested for its existence it was a truthiness test and was (apparently) not specific enough.

### DIFF
--- a/src/AutoComplete/AutoComplete.component.tsx
+++ b/src/AutoComplete/AutoComplete.component.tsx
@@ -16,39 +16,39 @@ type AutoCompleteDataSource = Array<{
 }>;
 
 interface AutoCompleteProps {
-    inputComponent?: any;
+    className?: string;
+    dataSource?: AutoCompleteDataSource | string[] | number[];
     debug?: boolean;
     name?: string;
-    dataSource?: AutoCompleteDataSource | string[] | number[];
-    className?: string;
-    size?: 'sm' | 'md' | 'lg';
     placeholder?: string;
+    size?: 'sm' | 'md' | 'lg';
     value?: string | number;
     // TODO: resultTemplate
     // TODO: allowClear
     allowClear?: boolean;
     // Visual
-    border?: boolean;
-    shadow?: boolean;
     background?: string;
+    border?: boolean;
     color?: string;
+    shadow?: boolean;
     // TODO: Allow children
     resultTemplate?: (props: any) => any;
-    suffix?: any;
+    inputComponent?: any;
     prefix?: any;
+    suffix?: any;
     // Event Handlers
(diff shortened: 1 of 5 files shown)


src/AutoComplete/AutoComplete.component.tsx
```diff
@@ -16,6 +16,7 @@ type AutoCompleteDataSource = Array<{
 }>;
 
 interface AutoCompleteProps {
+    inputComponent?: any;
```
zero0Halo: Following the pattern of other Anchor components that accept components as a prop. `ReactElement` throws TS errors, I'm guessing because of the `forwardRef`?

sharkparty: Probably, why not use the `as` prop built into styled-components? I think I'm not up to speed on the ask here, though. Wanna chat about that in person or dump it here?

src/AutoComplete/AutoComplete.component.tsx
```diff
@@ -16,38 +16,39 @@ type AutoCompleteDataSource = Array<{
...
-    border?: boolean;
-    shadow?: boolean;
     background?: string;
+    border?: boolean;
     color?: string;
+    shadow?: boolean;
     // TODO: Allow children
     resultTemplate?: (props: any) => any;
-    suffix?: any;
+    inputComponent?: any;
```
sharkparty, line 36: Still not clear on this prop. Have you seen [this](https://www.styled-components.com/docs/api#as-polymorphic-prop)?

zero0Halo: The `inputComponent` prop allows the user to provide their own JSX to replace the `Input` component being used. Theoretically any JSX could be supplied, but realistically it allows a new `Input` to be supplied with whatever attributes the user wants to tack on. Line `168` has an example of the usage.

I looked into `as` and I'm not exactly sure how that would solve this. I tried `<AutoComplete as="input" autoCapitalize="false />` and get TS errors so I'm guessing that's not what you're suggesting.

zero0Halo: I tried using `as` and passing the custom component into it. The problem is that the `ref` gets lost when doing so. I've tried a variety of ways to pass it along including the `forwardedAs` prop from `styled-components` with no luck.

PR #153, opened 2019-11-13
feat: ResponsiveProvider

zero0Halo:

---------

* Adds a new component, `ResponsiveProvider`. Put near the top of an app, this will measure the window's `innerWidth` and put it into context, along with sorted breakpoints read from the `ThemeProvider`. Most of the following components will tap into this. Also has a `debug` mode which shows the window's `innerWidth` and the current breakpoint in the bottom-right corner.

* `ResponsiveContext`, which `ResponsiveProvider` uses, is exported so that devs can easily access breakpoint data.

* The functionality noted above in `ResponsiveProvider` has been stripped out of `Grid`. `Grid` uses its own context to communicate `debug` mode to `Cell` children.

* `Cell` has been refactored to better use state. Aside from the improvements in readability, it also solves a bug in which the user would see a brief blip of the mobile view before seeing the correct layout for the current breakpoint. 

* `LegacyDesktop`, `LegacyTablet`, `LegacyPhone`, `LegacyBreakpoints`, `StandardBreakpoints` and `CenteredCell` have been removed as they no longer serve a purpose *in the new order*.

* `XS`, `SM`, `MD`, `LG`, `XL` and `XXL` components all use `ResponsiveProvider`. As with the other breakpoint implementations, internally these components don't care what the breakpoints are actually called, or even if there are 6 of them.

* `CustomAdaptor` has been updated to include `from` and `to` props. These are an abstraction for the user to access the breakpoint data, allowing them to pass `xs`, `sm` etc rather than a hard value.

* Bumped to version 1.3.0 as this is a significant difference from how responsive was handled previously.


src/Grid/Adaptor/Adaptor.component.tsx
```diff
@@ -27,144 +33,117 @@ interface FixedAdaptor extends CustomAdaptorProps {
...
+        <Responsive
+            className="anchor-adaptor"
+            {...props}
+            minWidth={min}
+            maxWidth={max}
+        />
+    ) : null;
+};
+
+export const XL = ({ ...props }: FixedAdaptor) => {
```
sharkparty: I don't know if these single breakpoint components are that useful, TBH. I'd like to explore making these better, but this re-solves what was here...

zero0Halo:
I've removed the `XS`, `SM`, `MD`, `LG`, `XL` and `XXL` components and renamed `CustomAdaptor` to `Adaptor`. Tests/stories updated. I also added a new field to `ResponsiveContext`, `current`, which gives the current breakpoint. Was a **Duh!** that I didn't have that already.

sharkparty (approved):
LGTM

PR #154, opened 2019-11-14
feat(Tooltip): upgrades, arrow and programatic show/hide

zero0Halo:

---------

My code does...

Adds the `Arrow` component from `DropDown` to tooltip and allows the user to override its default show/hide behavior by allowing them to pass a `hidden` prop which can be used to control it programatically.

### DIFF
--- a/src/Tooltip/Tooltip.component.tsx
+++ b/src/Tooltip/Tooltip.component.tsx
@@ -17,7 +17,7 @@ interface TooltipContainerProps extends React.HTMLAttributes<HTMLDivElement> {
     display?: string;
     maxWidth?: string;
     position?: Position;
-    hidden?: boolean;
+    show?: boolean;
     shadow?: string;
     wrapContent?: boolean;
     // Arrow props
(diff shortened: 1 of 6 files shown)


src/Tooltip/Tooltip.component.tsx
```diff
@@ -121,37 +155,48 @@ export class Tooltip extends React.Component<
...
             height,
             width,
-            hidden,
+            hidden : stateHidden,
             toolTipHeight,
             toolTipWidth,
         } = this.state;
+        // If the dev passes hidden as a prop then Tooltip's behavior changes, and whether or not it
+        // is visible is controlled by that prop and not by state.
+        const hiddenOveride = propsHidden !== undefined;
```
zero0Halo: I'm uncertain about this approach. I feel like it makes logical sense, but it feels somewhat tacked on.

brandonlilly: I think this is fine, but I would probably invert the logic to a `show` or `open` prop instead. It wouldn't be obvious to me that `hidden={false}` opens it up.

zero0Halo: I like that better. I had a feeling that maybe it was just the wording of the prop that was bugging me.

PR #157, opened 2019-11-19
fix: grid/cell media queries

zero0Halo:

---------

* Removes the `styled-css-grid` library as we now are rolling our own code for `Grid` and `Cell`.

* `Cell` now leverage's xstyled's `breakpoints()` function for responsiveness. A utility function reads all responsive information from passed props and concatenates their css and associate's it to a site's breakpoints. These are sorted and fed to the `StyledCell` component in order for mobile-first responsiveness.


src/Grid/Grid/Grid.component.spec.tsx
```diff
@@ -36,6 +43,22 @@ describe('Component: Grid & Cell', () => {
         expect(tree).toMatchSnapshot();
     });
 
+    // it('should correctly concatenate areas', () => {
```
sharkparty: We can prolly drop this test if it's not needed.

src/Grid/Grid/Grid.component.tsx
```diff
@@ -51,35 +50,69 @@ type AlignContent =
...
+        debug,
+        flow = FLOW.row,
+        gap = '0.5rem',
+        height = 'auto',
+        justifyContent,
+        minRowHeight = '1.25rem',
+        rowGap,
+        rows,
+    }) => css`
+        height: ${height};
```
sharkparty, line 90: I'm surprised you didn't use the object format for this. It all shakes out the same in the end, but this seems a lot more verbose. No need to change it—just commenting.

src/Grid/utils.ts
```diff
@@ -105,25 +82,98 @@ export function getBreakpointKey(
...
+    Small helper object for the generateBreakpointCSS() function to render the correct CSS for each
+    grid setting.
 */
-export function getResponsiveValue(
-    responsiveSettings: object | number | undefined,
-    innerWidth: number,
-    sortedBreakpoints: BreakpointType[]
+const ops = {
+    // If width is 0, don't show the cell. If middle is true, add in the middle CSS.
+    width: (w: any, middle: boolean) =>
```
sharkparty: This is probably copied code, but we should never self-minify vars. Unless we're coding in go where that's a language preference, we should be verbose because our peers will benefit from the verbosity and our build tools will transpile this down to single-letter vars.

src/Grid/utils.ts
```diff
@@ -105,25 +82,98 @@ export function getBreakpointKey(
...
+    const sortedResponsiveCSS: BreakpointType[] = [];
+    const generalSettings: GridSettings = {};
+
+    // Loops through the props with responsive settings, i.e. left, top, etc
+    Object.keys(gridSettings).forEach((settingKey: string) => {
+        const currentValue = gridSettings[settingKey];
+
+        // If the value passed isn't an object then it's not responsive, so its value should be
+        // put into the generalSettings object
+        if (typeof currentValue !== 'object') {
```
sharkparty: Fairly fragile type-check here. I don't think this will be an issue, but I figured I'd highlight it.

src/Grid/utils.ts
```diff
@@ -105,25 +82,98 @@ export function getBreakpointKey(
...
+            generalSettings[settingKey] = currentValue;
+        } else {
+            Object.keys(currentValue).forEach((breakpointKey: string) => {
```
sharkparty: O(n^2) here. This is a small object, but can we get this closer to linear time? Maybe via a frequency hash of both of these objects?

src/Grid/utils.ts
```diff
@@ -105,25 +82,98 @@ export function getBreakpointKey(
...
+                const responsiveValue = currentValue[breakpointKey];
+
+                if (responsiveCSS[breakpointKey]) {
+                    responsiveCSS[breakpointKey] += ops[settingKey](
```
sharkparty: This may be more succinct.

```
if (!responsiveCSS[breakpointKey]) {
    responsiveCSS[breakpointKey] = '';
}
responsiveCSS[breakpointKey] += ops[settingKey](
    responsiveValue,
    middle
);
```

src/Grid/utils.ts
```diff
@@ -105,25 +82,98 @@ export function getBreakpointKey(
...
+                    );
+                }
+            });
+        }
+    });
+
+    // This sorts the breakpoints by the breakpoints from ResponsiveProvider. This is neccessary
+    // because the order in which the media queries are rendered is important.
+    sortedBreakpoints.forEach((breakpointObj: BreakpointType) => {
+        const breakpointKey = Object.keys(breakpointObj)[0];
```
sharkparty: `const [breakpointKey,] = Object.keys(breakpointObj);`

sharkparty (changes requested):
I may be being a little too picky on this. Let me know if I need to help out with some of this. This is honestly a good PR and what I'm asking for is mostly enhancements that aren't necessarily essential, but should be on our radar. The quadratic runtime issue is glaring, but the qty of ops is likely so small it won't really hurt us or our users.

src/Grid/Grid/Grid.component.tsx
zero0Halo: No specific reason why I didn't. I'll update for consistency.

zero0Halo: Actually I take it back. Because most of the properties are optional I feel it's cleaner and simpler this way. I started converting it solely to an object with values and it was becoming much more verbose but for little benefit.

src/Grid/utils.ts
zero0Halo: Actually not copied. It was a small utility object/methods so I was continuing to keep it small. Point taken though, I'm normally a fan of variable verbosity.

src/Grid/utils.ts
zero0Halo: Fair. I was relying on the typing to only allow for 3 types (`undefined`, `number` and `BreakpointsType`) but it could be more explicit.

src/Grid/utils.ts
zero0Halo: Duh. I frequently forget you can deconstruct arrays.

sharkparty (approved):
This is all still executing in exponential time, but I don't want to block this PR any longer.

PR #161, opened 2019-12-04
fix: Cell now has default height and width of 1

zero0Halo:

---------

My code does...

### DIFF
--- a/src/Grid/Cell/Cell.component.tsx
+++ b/src/Grid/Cell/Cell.component.tsx
@@ -70,13 +70,13 @@ export class Cell extends React.PureComponent<CellProps> {
         super(props, context);
 
         // Generates css data for xstyle'd media queries in the constructor so as to only
-        // fire a single time and before render.
+        // fire a single time and before render. Width and height have a default of 1.
         const { sortedResponsiveCSS, generalSettings } = generateBreakpointCSS(
             {
                 left: props.left,
-                height: props.height,
+                height: props.height || 1,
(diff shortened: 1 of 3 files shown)


src/Grid/Cell/Cell.component.tsx
```diff
@@ -70,13 +70,13 @@ export class Cell extends React.PureComponent<CellProps> {
...
-        // fire a single time and before render.
+        // fire a single time and before render. Width and height have a default of 1.
         const { sortedResponsiveCSS, generalSettings } = generateBreakpointCSS(
             {
                 left: props.left,
-                height: props.height,
+                height: props.height || 1,
                 top: props.top,
-                width: props.width,
+                width: props.width || 1,
```
brandonlilly, line 79: What are the implications of the width defaulting to 1? Won't that mean by default the cell takes up 1/12 of the width of the grid? Is that the behavior that we want?

zero0Halo: Put it this way, it _was_ the intended behavior. The original library worked this way, and when I updated it for TypeScript I accidentally removed the default settings. 

The `Grid` itself has a default number of columns of 12, I think it's reasonable that cell's have a default width of 1. This is also easily overwritten via the `columns` prop in `Grid`. Its styles take precedence. For example, setting `columns` to `auto auto auto` still sets all `Cell`'s to auto width despite the default.

sharkparty (approved):
LGTM

PR #162, opened 2019-12-05
feat(Table): create table component

brandonlilly:

---------

Adds a basic Table component


src/Table/Table.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+    ${maxWidthStyles}
+    ${minWidthStyles}
+    ${paddingStyles}
+    ${fontSizeStyles}
+    ${fontWeightStyles}
+    ${colorStyles}
+    ${({ align }) => textAlignStyles({ textAlign: align })}
+`;
+
+export const TableHeaderCell = ({
```
brandonlilly, line 272: Another option is to nix the `Table.HeaderCell` component and add a prop for the `Cell` component to specify that its a `th` element.

src/Table/Table.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+    ${widthStyles}
+    ${maxWidthStyles}
+    ${minWidthStyles}
+    ${({ borderRadius = 'base' }) =>
+        css({
+            borderRadius,
+        })}
+
+    tbody tr:nth-child(odd) {
+        background-color: background.base;
```
brandonlilly: Perhaps there should be a prop for the alternate background color.

src/Table/Table.component.tsx
sharkparty: I _think_ this is more semantic. 🤔

src/Table/Table.component.tsx
sharkparty: 100% on this

src/Table/Table.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+        MarginProps,
+        WidthProps,
+        MaxWidthProps,
+        MinWidthProps {
+    borderRadius?: string;
+    size?: string;
+}
+
+const StyledTable = styled('table')<TableProps>`
+    border: base;
```
sharkparty: Shouldn't this be none? Or a prop?

src/Table/Table.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+}: TableHeaderCellProps): React.ReactElement<TableHeaderCellProps> => (
+    <StyledTh
+        className={classNames('anchor-table-header-cell', className)}
+        {...props}
+    >
+        {children}
+    </StyledTh>
+);
+
+Table.HeaderCell = TableHeaderCell;
```
sharkparty, line 285: Nice choice on these exports. I typically try to have them all in independent files, but given our chats about this component, this really sets us up perfectly to scale what we need to and throw away what we don't without a bunch of useless boilerplate code.

src/Table/Table.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+        const component = shallow(subject);
+
+        expect(subject).toBeDefined();
+        expect(wrapper).toBeDefined();
+        expect(component).toBeDefined();
+        const tree = renderer.create(subject).toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+
+    it('should render with all the bells and whistles', () => {
```
sharkparty, line 36: I don't love the kitchen sink approach here, but as these are mostly pass-throughs it seems ok.

src/Table/Table.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+import * as renderer from 'react-test-renderer';
+import { ThemeProvider } from '@xstyled/styled-components';
+import { shallow, mount } from 'enzyme';
+// ANCHOR
+import { RootTheme } from '../../src/theme';
+import { Table } from './Table.component';
+const { Head, Body, Cell, Footer, Row, HeaderCell } = Table;
+
+describe('Component: Table', () => {
+    it('should be defined.', () => {
```
sharkparty: `'should match snapshot'`. You can hoist the expect/tobedefined to the root describe block like I did with the widget.

src/Table/Table.component.tsx
zero0Halo: 200% this

src/Table/Table.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+    tr {
+        border: dark;
+    }
+
+    th {
+        white-space: nowrap;
+    }
+`;
+
+StyledTable.defaultProps = { borderRadius: 'base' };
```
zero0Halo: Is this necessary? You've declared 'base' as the default on line 84.

src/Table/Table.component.tsx
zero0Halo: Somewhat related; perhaps the `Cell` in `Grid` should follow this pattern to be consistent. I don't generally like name-spacing the components but `Cell` for `Table` and `Cell` for `Grid`, it's having me rethink my position.

src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+                                color="text.light"
+                                weight="normal"
+                            >
+                                (90 Day Qty)
+                            </Typography>
+                        </HeaderCell>
+                    </Row>
+                </Head>
+                <Body>
+                    {data.map(({ name, color, rating }) => (
```
zero0Halo, line 82: Since you're mapping you need a key.

src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+                                scope="row"
+                                align="left"
+                                color="#91609c"
+                                fontWeight={600}
+                            >
+                                {name}
+                            </HeaderCell>
+                            <Cell
+                                align="center"
+                                width={{ xs: '100px', md: '1px' }}
```
zero0Halo: I think it's good to demonstrate what sets this `Table` apart from using a standard `table`. Perhaps set width's here to something really noticeable, because `1px` doesn't actually do anything since table cells can't totally collapse, and `xs` is difficult to see in Storybook right off the bat.

src/Table/Table.component.tsx
zero0Halo: I would make `border` a prop, but keep `base` as the default.

src/Table/Table.stories.tsx
zero0Halo: The real benefit of this implementation of `table` is all the xstyled system props you baked in. I'd consider making more knobs to show off just how configurable this is.

src/Table/Table.component.tsx
sharkparty: This is a hill I won't die on. I like the dedication to consistency and I'm happy to go either way for the sake of what I'll dub "strict consistency" or "moderate consistency."

brandonlilly: I think the grid Cell is more fundamental, and I'm fine with it not being namespaced.

zero0Halo (approved):
I would have liked to have seen you tweak the `width` values for the `Cell` in the stories, but other than that this looks good!

PR #172, opened 2019-12-19
fix: rename typings file extension to be .ts

mintyfloss:

---------

This fixes errors that I'm seeing

```
[typescript] node_modules/@retailmenot/anchor/ts/theme/index.d.ts(1,23): error TS2688: Cannot find type definition file for 'typings'.
[typescript] node_modules/@retailmenot/anchor/ts/utils/position/position.d.ts(1,23): error TS2688: Cannot find type definition file for 'typings'.
```

In `position.d.ts`, there's a reference to `typings` but it doesn't exist.

```
/// <reference types="typings" />
export declare type Position = 'topStart' | 'top' | 'topEnd' | 'rightStart' | 'right' | 'rightEnd' | 'bottomEnd' | 'bottom' | 'bottomStart' | 'leftEnd' | 'left' | 'leftStart';
export declare const positionVariants: (position: Position, height: number, width: number, overlayHeight: number, overlayWidth: number, spacing?: number) => import("styled-components").FlattenSimpleInterpolation;
```

The problem is that the `typings.d.ts` isn't being copied over to the `ts` folder during a build. It looks like that is by design:

https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build
https://stackoverflow.com/questions/16687340/tsc-out-declaration-is-not-including-custom-interfaces

I'm not sure if this is the best solution, but I'm following the suggestions in the two links above and renaming `typings.d.ts` to `typings.ts`.


sharkparty:
@mintyfloss — do you know what version you're on? We do ship the typings file. We use the manual copy step.

sharkparty (approved):
I spent some time coming up to speed and now I think I see what you're talking about. I did build this locally and it presented a new problem, but I'm curious if this solves your current one. Let's give it a shot.

PR #175, opened 2020-01-15
Adding Space Props

sharkparty:
added spacing props to top-level comps; and,
refactored naming convention

---------

My code removed padding from ListItems in a List; added spacing props to top-level comps; and, refactored naming convention


src/Typography/Typography.component.tsx
```diff
@@ -139,13 +139,13 @@ const StyledTypography = (tag: TypographyTags) => styled(tag)<
 export const Typography = ({
     className,
     children,
-    tag = 'span',
+    tag: as = 'span',
     color,
     ...props
 }: TypographyProps): JSX.Element =>
-    React.createElement(StyledTypography(tag), {
+    React.createElement(StyledTypography(as), {
```
brandonlilly: Is `React.createElement` necessary anymore with using the `as` prop?

Also this line above
```tsx
const StyledTypography = (tag: TypographyTags) => styled(tag)<
```

src/Table/Table.component.tsx
```diff
@@ -64,7 +66,8 @@ interface TableProps
         MarginProps,
```
brandonlilly: `MarginProps` are redundant now (same with `marginStyles` below)

src/Modal/Footer/__snapshots__/Footer.component.spec.tsx.snap
```diff
@@ -23,6 +23,7 @@ exports[`Component: Modal.Footer should match snapshot with padding 1`] = `
   order: 1;
   justify-self: flex-end;
   background: transparent;
+  padding: 1rem;
```
brandonlilly, line 26: Lol

src/Alert/__snapshots__/Alert.component.spec.tsx.snap
```diff
@@ -48,7 +48,7 @@ Array [
   box-sizing: border-box;
   margin: 0;
   font-family: Avenir Next,Segoe UI,Helvetica Neue,Helvetica,Roboto,sans-serif;
-  font-size: 1rem;
+  font-size: 16px;
```
brandonlilly: How did this happen?

src/Tabs/Tab/Tab.component.tsx
```diff
@@ -2,11 +2,11 @@
 import * as React from 'react';
 import classNames from 'classnames';
 import styled, { css, ThemeContext } from '@xstyled/styled-components';
-import { th, space as spaceStyles } from '@xstyled/system';
+import { th, space as spaceStyles, SpaceProps } from '@xstyled/system';
 // ANCHOR
 import { Position, Variant, variantStyles } from '../utils';
 
-interface TabProps extends React.HTMLAttributes<HTMLDivElement> {
+interface TabProps extends React.HTMLAttributes<HTMLDivElement>, SpaceProps {
```
brandonlilly, line 9: These are redundant now:
```tsx
    margin?: string;
    padding?: string;
```

src/Modal/Modal.component.tsx
```diff
@@ -68,9 +68,10 @@ const StyledModal = StyledReactModal.default.styled`
         min-height: ${({ size = defaultSize }: ModalProps) =>
             Sizes[size].footerHeight}rem;
     }
+    ${spaceStyles}
```
brandonlilly, line 69: This ones a little more complicated with the 
```tsx
    margin: ${({ margin = 'auto' }: ModalProps) => margin};
```
above

src/Typography/Typography.component.tsx
sharkparty: I was aiming for a partial refactor here, but it may make more sense for this to be a breaking change. Styled-Comps has supported a global `as` prop since V4 and we just don't need this anymore.

sharkparty: https://styled-components.com/docs/api#as-polymorphic-prop

src/Alert/__snapshots__/Alert.component.spec.tsx.snap
sharkparty: I'm not sure, but the updated one isn't in px.

src/Alert/__snapshots__/Alert.component.spec.tsx.snap
```diff
@@ -2,7 +2,22 @@
 
 exports[`Component: Alert should match its snapshot. 1`] = `
 Array [
-  .c0 {
+  .c1 {
+  box-sizing: border-box;
+  margin: 0;
+  font-family: Avenir Next,Segoe UI,Helvetica Neue,Helvetica,Roboto,sans-serif;
+  font-size: 1rem;
```
sharkparty, line 9: Yay `rem`s! @brandonlilly

package.json
```diff
@@ -41,9 +41,8 @@
     "build:modules": "rollup -c rollup.config.js",
     "ci": "concurrently --names 'lint,test' 'npm run lint' 'npm run test'",
     "prepublishOnly": "npm run build",
-    "lint": "concurrently --names 'pretty,tslint' 'npm run pretty' 'npm run tslint'",
```
brandonlilly, line 44: No more lint?

src/theme/typography.theme.ts
```diff
@@ -115,6 +115,7 @@ export const typography: TypographyTheme = {
         h1: {
             fontSize: '1.75rem',
             lineHeight: '2rem',
+            color: 'red',
```
brandonlilly: ^

src/Typography/Typography.component.tsx
```diff
@@ -139,14 +137,14 @@ const StyledTypography = (tag: TypographyTags) => styled(tag)<
 export const Typography = ({
     className,
     children,
-    tag = 'span',
+    as = 'span',
```
brandonlilly: I think specifying the default here should be redundant, since its a `span` normally. The only difference afaict would be that the prop isn't defined by default for the variant styles- but we can specify a default there anyway (which is currently set to `none`).

src/Typography/Typography.component.tsx
```diff
@@ -139,14 +137,14 @@ const StyledTypography = (tag: TypographyTags) => styled(tag)<
 export const Typography = ({
     className,
     children,
-    tag = 'span',
```
brandonlilly, line 142: If we want to deprecate over time, we could accept `tag` pass it in to `as` and console.warn that `tag` is being deprecated.

src/Typography/Typography.component.tsx
```diff
@@ -139,14 +137,14 @@ const StyledTypography = (tag: TypographyTags) => styled(tag)<
...
-        children,
-        tag,
-        $color: color,
-        className: classNames('anchor-typography', className),
-        ...props,
-    });
+}: TypographyProps): JSX.Element => (
+    <StyledTypography
+        className={classNames('anchor-typography', className)}
+        {...{ as, color, ...props }}
```
brandonlilly: `$color: color` keeps `color="text.base"` out of the dom.

you should be able to keep it out of this object spread and just do `$color={color}`

src/List/Title/__snapshots__/Title.component.spec.tsx.snap
```diff
@@ -24,19 +10,34 @@ exports[`Component: Title should be defined 1`] = `
   font-weight: 500;
   text-align: inherit;
   color: inherit;
-  color: #808080;
```
brandonlilly: This shouldn't be gone (see `$color` vs `color` comment)

src/List/Title/__snapshots__/Title.component.spec.tsx.snap
```diff
@@ -24,19 +10,34 @@ exports[`Component: Title should be defined 1`] = `
...
+  height: 2rem;
+  margin-top: 0.75rem;
+}
+
 <div
   className="c0 anchor-list-title"
 >
   <span
     className="c1 anchor-typography"
+    color="text.label"
```
brandonlilly: But this should be gone (see `$color` vs `color` comment)

src/AutoComplete/ResultsContainer/__snapshots__/ResultsContainer.component.spec.tsx.snap
```diff
@@ -363,7 +290,8 @@ exports[`Component: ResultsContainer should be defined 1`] = `
       onMouseOver={[Function]}
     >
       <span
-        className="c5 anchor-typography c4 lg"
```
brandonlilly, line 366: I'm confused by why `anchor-typography` is disappearing.

package.json
sharkparty: So tslint as a project is EOL'd, but lemme add eslint with the ts plugins.

src/theme/typography.theme.ts
sharkparty: Nice catch.

src/Typography/Typography.component.tsx
sharkparty: I was initially doing just that, but this has been included in the SC lib since v4 and we really just need to move this forward.

src/Typography/Typography.component.tsx
sharkparty: Another artifact from my naive first SC.

src/List/Title/__snapshots__/Title.component.spec.tsx.snap
sharkparty: It is now, thanks!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@retailmenot/anchor",
-  "version": "1.3.7",
+  "version": "1.4.0",
```
jcristol, line 3: just curious why jump a new minor version and a couple of bug fix versions?

src/utils/position/position.ts
```diff
@@ -20,7 +20,7 @@ export const positionVariants = (
     width: number,
     overlayHeight: number,
     overlayWidth: number,
-    spacing: number = 0
+    spacing = 0
```
jcristol, line 23: was this removed for redundancy sake?

src/Tabs/Tabs.component.tsx
```diff
@@ -94,12 +94,12 @@ export const Tabs = ({
     onChange,
     ...props
 }: TabsProps): React.ReactElement<TabsProps> => {
-    let tabInfo: Array<{
```
jcristol, line 97: why move from `Array` of blah to this other notation?

jcristol (commented):
Added some review items mostly for my education. Skimmed through the Components because they all look like they similar change sets. But as a whole this makes sense.

package.json
sharkparty: This introduces breaking changes with the Typog..y component, but it's merely a consistency fix with a lower level dependency so I felt like a good middle-ground would be a minor release.

src/utils/position/position.ts
sharkparty: This is a lint rule I enabled. The tl;dr: is do not allow people to specify a type redundantly when it can be inferred from the instantiation.

src/Tabs/Tabs.component.tsx
sharkparty: Hah, I actually kinda hate this. It's another lint rule that I enabled. While I hate this, the Typescript community is more widely swayed by the `[]int` syntax than `Array<int>` and the most compelling reason is that `Array` is a JS language construct that is being used by TS as an interface which is "confusing". I'm cool switching this back because I think it reads better, but this seems to be more widely adopted by the TS community, especially those in project we borrow capital from.

jcristol (approved):

jcristol (approved):

jcristol (approved):

jcristol (approved):

src/AutoComplete/ResultsContainer/__snapshots__/ResultsContainer.component.spec.tsx.snap
```diff
@@ -156,7 +137,7 @@ exports[`Component: ResultsContainer should accept a resultTemplate 1`] = `
       onMouseOver={[Function]}
     >
       <span
-        className="c2 anchor-typography c3 lg"
+        className="sc-LzLte c2 lg anchor-typography"
```
brandonlilly, line 140: The change I made to the Typography components isn't playing 100% nice with the `jest-styled-components` render. Afaict, all of where this occurs has the same hash (`sc-LzLte`). The only real potential issue in my mind is test fragility (i.e. a simple change modifies the hash, requiring a bunch of snapshot update).

src/Typography/Typography.component.tsx
```diff
@@ -88,9 +86,14 @@ export interface StyledTypographyProps extends TypographyProps {
     $color?: 'inherit' | 'initial' | string;
 }
 
-const StyledTypography = (tag: TypographyTags) => styled(tag)<
-    StyledTypographyProps
->`
+export const Typography = styled('span').attrs<TypographyProps>(
```
brandonlilly, line 89: Converting Typography to a `styled` component (rather than a component that returns a styled component) allows styled-components to squash the styles of components that styled-extend Typography (i.e.  styled(Typography)).

The issue with the missing `anchor-typography` classes was with components that both extend Typography (e.g. Hero.Title, Alert, List.Item) and use the `as` prop. The explanation is complicated (https://github.com/styled-components/styled-components/issues/1981#issuecomment-419287003) and is also the expected behavior for styled-components. This change fixes that issue.

sharkparty: I read through this and it's a known issue and while I don't like that we lose the ability to access typography by it's class under certain conditions, I don't think this is a gate for this PR.

sharkparty: Also, thanks for finding this and spending the time to explain it.

brandonlilly (approved):
LFTM

PR #177, opened 2020-01-23
feat: utility method, themeMerge

zero0Halo:

---------

* Adds our first utility method, `themeMerge` as part of the `theme` export. Allows the user to merge an object into the `RootTheme` which returns a new object (deep clone).


src/theme/themeMerge/themeMerge.ts
```diff
@@ -0,0 +1,69 @@
...
+            tmpObj[key] = deepClone(data[key]);
+        }
+    }
+    return tmpObj;
+}
+
+export function themeMerge(settings: object, rootTheme: ThemeType): ThemeType {
+    const newTheme = deepClone(rootTheme);
+
+    const recursiveUpdate = (settingStep: object, newThemeStep: object) => {
```
brandonlilly: Looks like this function can be moved outside of the `themeMerge` definition.

src/theme/themeMerge/themeMerge.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+import { RootTheme } from '../index';
+
+// TEST SETUP
+const data = {
+    fonts: {
+        base: 'Comic Sans MS',
+    },
+    colors: {
+        primary: {
+            base: '#ff0000',
```
brandonlilly: Can we add some more test cases for more complicated edgecases?

I'm thinking things like:
```ts
    exampleComponent: {
        base: ({ color }) => css`
            color: ${color};

        `,
        dark: css`
            color: ${props => props.color};
            background: black;
        `,
    },
```

I'd also love to see a test case that actually uses the resultant theme in a theme provider and verifies the theme styles being applied to a component in a snapshot.

brandonlilly: What about if someone wants to remove an entry from our theme, e.g.

```ts
themeMerge({
  buttons: {
    filled: undefined,
  },
});
```

.github/OWNERS
```diff
@@ -39,5 +39,6 @@
 /src/Form/ControlLabel @brandonlilly
 /src/Table @brandonlilly
 /src/Tabs @brandonlilly
+/src/theme/themeMerge @zero0Halo
```
brandonlilly: What do you think of moving the file from `/src/theme/themeMerge` to `/src/utils/themeMerge`?

src/theme/index.ts
```diff
@@ -43,3 +44,7 @@ export const RootTheme = {
 
 // Exporting out @xstyled's ThemeProvider. Only way I could get it to work w/o issues.
 export class ThemeProvider extends XstyledThemeProvider {}
+
+// Passing RootTheme to avoid circular references with exports
+export const themeMerge = (settings: object) =>
```
brandonlilly: Can you make it so that `RootTheme` is optional here, so that consumers can more easily create multiple subthemes?

src/theme/themeMerge/themeMerge.ts
```diff
@@ -0,0 +1,69 @@
...
+    buttons: object;
+    inputs: object;
+    colors: object;
+    borders: object;
+    skeleton: object;
+    [key: string]: any;
+};
+
+// Can't really be 100% sure what data will be as it is cloned, hence :any
+function deepClone(data: any) {
```
brandonlilly: Can we get at least 1 test for this function, to verify that mutating the clone doesn't mutate the original?

.github/OWNERS
zero0Halo: Not opposed. I put in with `RootTheme` and `ThemeProvider` because they are all related, but as it's not a component I think it's fine. I can probably get around the circular logic I was encountering by moving it out of the same export, too.

src/theme/themeMerge/themeMerge.ts
zero0Halo: This can be accomplished by testing `themeMerge` itself, I neglected to write a test for it though which was dumb.

src/theme/themeMerge/themeMerge.spec.ts
zero0Halo: Definitely will add some snapshot tests.

Someone wanting to remove an entry I think is a bigger conversation and out of the scope of this initial offering. We've been pushing hard for people _not_ to do that since they're still learning how to use theming. Maybe something to address later.

zero0Halo: Added a snapshot test for the `Button` as that has edge-case you defined.

.github/OWNERS
brandonlilly: File location changed, so this should update too

brandonlilly (commented):
```tsx
    exampleComponent: {
        base: ({ color }) => css`
            color: ${color};

        `,
        dark: css`
            color: ${props => props.color};
            background: black;
        `,
    },
```

I'm not seeing any examples with the more complicated use cases ^

src/theme/index.ts
```diff
@@ -43,3 +45,10 @@ export const RootTheme = {
 
 // Exporting out @xstyled's ThemeProvider. Only way I could get it to work w/o issues.
 export class ThemeProvider extends XstyledThemeProvider {}
+
+// Passing RootTheme to avoid circular references with exports. RootTheme is the default baseTheme
+// but the user can pass in any theme.
+export const themeMerge = (
```
brandonlilly: Is there still a need to do this and export from the theme file, as opposed to exporting the function directly from the utils file?

zero0Halo: None of the util functions are exports for the main Anchor build. If we want to start doing that, we can. I have no strong opinion.

src/utils/themeMerge/themeMerge.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+            </ThemeProvider>
+        );
+
+        const tree = renderer.create(<Test />).toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+
+    it('should not affect values set with the css() function', () => {
+        const Test = () => (
+            <Tabs variant="tabular">
```
brandonlilly: This should the `regular`

brandonlilly (approved):
Just one thing

PR #182, opened 2020-02-03
chore: add a trigger type argument to DropDown onTrigger

reedham1:

---------
- Added string parameter to the onTrigger function to specify the type of trigger (so parent components can build custom logic around dropdown)
- Unit tests

### DIFF
--- a/src/DropDown/DropDown.component.tsx
+++ b/src/DropDown/DropDown.component.tsx
@@ -24,7 +24,7 @@ interface DropDownProps extends React.HTMLAttributes<HTMLDivElement> {
     border?: string;
     borderRadius?: string;
     debug?: boolean;
-    onTriggered?: (state: boolean) => void;
+    onTriggered?: (state: boolean, trigger?: string) => void;
     position?: Position;
     ref?: any;
     shadow?: string;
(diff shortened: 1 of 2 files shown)


src/DropDown/DropDown.component.tsx
```diff
@@ -24,7 +24,7 @@ interface DropDownProps extends React.HTMLAttributes<HTMLDivElement> {
     border?: string;
     borderRadius?: string;
     debug?: boolean;
-    onTriggered?: (state: boolean) => void;
+    onTriggered?: (state: boolean, trigger?: string) => void;
```
sharkparty, line 27: This is something we can unit test, and should. What's the end-goal of changing this util method?

reedham1: We have a button in a navbar that you click/hover to trigger the dropdown, and we want the button to respond to the dropdown (color change, chevron rotate, etc). Wanting to use `onTriggered` to track the state of the dropdown, but it is not useful when using the 'both' value for the `trigger` prop.

Happy to write unit tests if you're ok with the change

reedham1: If the button knows which trigger type is firing `onTriggered`, we can write a bit of logic to respond appropriately

sharkparty (approved):
LGTM

PR #186, opened 2020-02-08
feat: allows evergreen documentation

zero0Halo:
chore(package.json): new script, build:for-docs

chore(docs/package.json): local reference to anchor build

fix(docs/Page, docs/index): ThemeProvider pulled from xstyled

---------

**What?**
Makes the `@retailmenot/anchor` reference in the `docs` site point to a local build, making it possible for the doc site to be synced with Anchor.

**How?**
`yarn build` works as before, except as a last step it also calls a new script, `build:for-docs`. This copies the `esm`, `commonjs`, `ts` folders, and `package.json` into an untracked folder, `for-docs`. Running `yarn` within the `docs` folder will then pull in this local build of Anchor from the `for-docs` folder.

**Why?**
I tried a lot before landing on this solution. Trying to locally add Anchor from docs did not work without the intermediate step of copying the npm files to the `for-docs` folder. Gatsby would always choke and die because the entire contents of Anchor would be imported rather than just the necessary `esm`, `commonjs`, `ts` and `package.json`. 

I tried modifying our build process in Anchor to output to a `build` folder and instead pull that in, but that approach doesn't work because `package.json` would still need to exist in that `build` folder, and the pathing would get messed up (i.e. within Anchor it would correctly point to `/build/esm/` etc, but within the `build` folder the copied `package.json` would not have a `build/esm` path.).

Creating an untracked `for-docs` lets us keep the current build process and make everything transparent. The only step to remember to take is when done working on a component, running `yarn build` before going into `docs` and running `gatsby develop/build`.

**Gotchas**
The lint step on Travis is failing because the `for-docs` folder is untracked which causes the `yarn build` process w/in the `docs` folder to fail. Adding steps to run `build` from Anchor first should take care of that, I think.

The only other one I've run into so far is that somehow this local reference is causing some small TypeScript highlight warnings/errors w/in the doc site. They're very minor and it doesn't prevent building. With a little more time, we can probably sort that out.


zero0Halo:
Fixed issue with Travis build failing.

zero0Halo:
Steps to try this out for the first time:

    cd anchor
    yarn build
    cd docs
    yarn
    gatsby develop

Then if you want make a change to an Anchor component, to see it reflected on the doc site:

    yarn build
    cd docs
    rm -rf node_modules
    yarn
    gatsby develop

Because this local folder is treated like any other package, that means that if you want to make a change to Anchor you'll have to either upgrade the version number, or delete the `@retailmenot/anchor` folder in the doc site's `node_modules` folder.

brandonlilly (approved):
Neat!

PR #192, opened 2020-02-24
feat(Adapt): create Adapt component

brandonlilly:

---------

Adds an Adapt component


src/Adapt/Adapt.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+ *       {value}
+ *     </p>
+ *   )}
+ * </Adapt>
+ *
+ * =>
+ *   <p>apple</p>  // displays below md breakpoint
+ *   <p>banana</p>  // displays at and above md breakpoint
+ */
+export const Adapt = ({
```
brandonlilly, line 74: `Adapt` name can be changed if we think of something better

src/Adapt/Adapt.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+    className,
+    children,
+    forEach,
```
brandonlilly, line 77: `forEach` I'm also not completely sold on

src/Adapt/Adapt.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+    [k: string]: any;
+}
+
+interface AdaptProps
+    extends React.HTMLAttributes<HTMLDivElement>,
+        LayoutProps,
+        SpaceProps {
+    className?: string;
+    forEach: ForEach;
+    children: (value: any, breakpoint: string) => any;
```
sharkparty: Won't this always need to return `JSX` or `null`?

src/Adapt/Adapt.component.tsx
sharkparty: I'm not usually the name police, but `target` or something along those lines is better than using a protected JS namespace IMO.

src/Adapt/Adapt.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+}: AdaptProps): React.ReactElement<AdaptProps> => {
+    const theme = useContext(ThemeContext);
+    const breakpoints = theme && theme.breakpoints;
+
+    const keys = Object.keys(forEach);
+
+    // keys are sorted by their pixel values (if available) so
+    // that the css styles are ordered properly
+    const sortedKeys = breakpoints
+        ? keys.sort((a, b) => (breakpoints[a] || 0) - (breakpoints[b] || 0))
```
sharkparty, line 89: Could also be `([a, ], [b,]) => (a || 0 ) - (b || 0 )...`

sharkparty: Actually, I read `keys` wrong. nvm

sharkparty (approved):

src/Adapt/Adapt.component.tsx
brandonlilly: I imagined the scenario that someone just passes in a string- but you're right, that wouldn't work with this system (need a dom node to target).

The possibility of `null` possibility is also very interesting. May need to deal with that specifically

src/Adapt/Adapt.component.tsx
brandonlilly: I avoided `for` and `each`, and while `forEach` is an array method, I wouldn't consider it protected namespace. The thing I like about it is the implication that the render prop (`children`) will render with every specified breakpoint.

```jsx
eachBreakpoint={{ ... }}
```
?

src/Hero/__snapshots__/Hero.component.spec.tsx.snap
```diff
@@ -213,7 +213,7 @@ exports[`Component: Hero should render with title 1`] = `
   className="c0 anchor-hero"
 >
   <h1
-    className="sc-qQYBZ c1 anchor-hero-title anchor-typography"
+    className="sc-qZtCU c1 anchor-hero-title anchor-typography"
```
brandonlilly, line 216: Just a note: These didn't break until I modified `src/index.ts`

src/Adapt/Adapt.component.tsx
sharkparty: What about `onBreakpoint`?

src/Adapt/Adapt.component.tsx
sharkparty: I think `any` is prolly fine for now, but whenever I see that in TS I assume it's incorrect.

src/Adapt/Adapt.component.tsx
zero0Halo: Since the syntax is so similar to xstyled's utility, what about `breakpoints`? Keeps it consistent and the API is (basically) the same.

src/Adapt/Adapt.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+    ...props
+}: AdaptProps): React.ReactElement<AdaptProps> => {
+    const theme = useContext(ThemeContext);
+    const breakpoints = theme && theme.breakpoints;
+
+    const keys = Object.keys(forEach);
+
+    // keys are sorted by their pixel values (if available) so
+    // that the css styles are ordered properly
+    const sortedKeys = breakpoints
```
zero0Halo, line 88: A suggestion: `ResponsiveProvider` already sorts the breakpoints and makes them available via `ResponsiveContext`. You could pull that in instead of `ThemeContext` and not have to sort the breakpoints for each instance of `Adapt`.

zero0Halo (commented):
Do we want to retire `Adaptor`? Having two components with such similar names and functions, feels confusing.

src/Adapt/Adapt.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+                <StyledStory>
+                    <Adapt
+                        forEach={object('forEach', {
+                            xs: 'apple',
+                            md: 'strawberry',
+                            lg: 'blueberry',
+                        })}
+                    >
+                        {(value, breakpoint) => (
+                            <div>
```
zero0Halo: Going to want to add a `key` prop here, and for the other example. Getting fussed at in the console.

src/Adapt/Adapt.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+                <Adapt
+                    forEach={{
+                        xs: 'xs',
+                        sm: 'sm',
+                        md: 'md',
+                        lg: 'lg',
+                    }}
+                >
+                    {(size, breakpoint) => (
+                        <Button size={size} margin="0">
```
zero0Halo: `<Button size={size} margin="0" key={breakpoint}>`

src/Adapt/Adapt.component.tsx
brandonlilly: I'll add a comment about it for now

src/Adapt/Adapt.component.tsx
brandonlilly: Just gonna keep `forEach` :)

PR #194, opened 2020-03-11
fix(Collapse): stop collapse icon from wrapping under button text

speed-rmn:

---------

Because the floated button icon came after the button text, if the icon needed to wrap, it would wrap under the text.  The solution is either to move the floated icon to be the first child of the button or to go with a flex-based approach.  Making the parent button a flexbox allows the icon to come second, which is more intuitive, and the default `align-items: stretch` rule means that all of the vertical space beneath the icon will be reserved, so button text will never wrap under the icon.  Additional tweaks were made to appearance and background-color, because in browsers other than Chrome a default "buttonFace" color (grey) was still showing through on this control.

Some docs were also updated for clarity or formatting purposes.

An issue with `isOpen` was identified in `CollapseGroup` with a TODO.

### DIFF
--- a/docs/src/pages/components/collapse.mdx
+++ b/docs/src/pages/components/collapse.mdx
@@ -66,7 +66,8 @@ The **`CollapseGroup`** component combines multiple **`Collapse`** components to
     },
     {
         property: 'closedText',
-        description: 'The text that appears on the button as a CTA when the component is closed.',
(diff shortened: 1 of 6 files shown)


brandonlilly (approved):
Nice 👍

speed-rmn:
Can't push new commits or force push squashed commits.  Created https://github.com/RetailMeNot/anchor/pull/195 to fix linting issue.

PR #198, opened 2020-04-01
fix(Button): support flip for cirulcar buttons

mintyfloss:

---------

My code does...

### DIFF
--- a/src/Button/Flip/Flip.component.tsx
+++ b/src/Button/Flip/Flip.component.tsx
@@ -1,24 +1,35 @@
 // VENDOR
 import * as React from 'react';
-import styled from '@xstyled/styled-components';
-import { th } from '@xstyled/system';
+import styled, { css } from '@xstyled/styled-components';
+
 // ANCHOR
 import { Theme, TRANSITION_SPEED, REVEAL_BACKGROUND_COLOR } from '../utils';
 
 export interface StyledFlipProps extends React.HTMLAttributes<HTMLDivElement> {
+    circular?: boolean;
     flipColor: string;
 }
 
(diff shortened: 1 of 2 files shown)


mintyfloss:
I have a snapshot to update, but I'm not able to push any changes :(

```
remote: error: GH006: Protected branch update failed for refs/heads/circular-flip.
remote: error: Required status check "continuous-integration/travis-ci" is expected. At least 1 approving review is required by reviewers with write access.
To github.com:RetailMeNot/anchor.git
 ! [remote rejected] circular-flip -> circular-flip (protected branch hook declined)
error: failed to push some refs to 'git@github.com:RetailMeNot/anchor.git'
```

src/Button/Flip/Flip.component.tsx
```diff
@@ -1,25 +1,36 @@
...
     top: -1px;
     right: -1px;
-    width: 1.25rem;
-    height: 1.25rem;
     content: '';
 
+    ${({ circular }) =>
+        circular
+            ? css`
+                  border-top-right-radius: 1000px;
```
brandonlilly, line 22: if `base` works below, then `circular` should work here.

mintyfloss: Ah yes, I didn't realize `circular` was in the theme. It looks like I have more work to do in this PR. Circular flip only works in the `md` size. The problem is that I want to make the flip be 50% of the button height and make it a square. I haven't thought of a good way to do that yet.

brandonlilly: make it a square on a circular button? do you have a mock?

`height` is something you can pass into the flip component from the button component, then just `calc(${height} / 2)` should work I think

PR #201, opened 2020-04-07
fix(Skeleton): support multiple color schemes

brandonlilly:

---------

Static keyframe names didn't play nice with multiple different color schemes existing at the same time. This switches it to use the styled-components `keyframes` api to support multiple different color schemes, and also enables skeleton variants.

### DIFF
--- a/src/Skeleton/Skeleton.component.tsx
+++ b/src/Skeleton/Skeleton.component.tsx
@@ -101,6 +101,9 @@ export const Skeleton = ({
     children,
     display,
     textLength,
+    // This is the unicode block character that put together looks like one
+    // long bar. There are shorter and taller characters that could be used
+    // or a fatter or thinner bar.
     textChar = '▆',
     // This is the conversion from block width to the average width of
     // a normal distribution of letters. We came up with it by eyeballing it
--- a/src/Skeleton/Skeleton.component.tsx
+++ b/src/Skeleton/Skeleton.component.tsx
@@ -1,70 +1,92 @@
-// REACT
-import * as React from 'react';
 // VENDOR
+import * as React from 'react';
 import classNames from 'classnames';
-import styled, { css } from '@xstyled/styled-components';
-import { th, space as spaceStyles, SpaceProps } from '@xstyled/system';
+import styled, {
+    css,
+    keyframes,
+    ThemeContext,
+} from '@xstyled/styled-components';
+import {
+    space as spaceStyles,
+    SpaceProps,
+    layout as layoutStyles,
+    LayoutProps,
+    borderRadius as borderRadiusStyles,
+    BorderRadiusProps,
+    variant as createVariant,
+    th,
+} from '@xstyled/system';
 
 interface SkeletonProps
     extends React.HTMLAttributes<HTMLDivElement>,
-        SpaceProps {
+        SpaceProps,
+        LayoutProps,
+        BorderRadiusProps {
     className?: string;
     loading?: boolean;
-    borderRadius?: string;
-    display?: string;
-    height?: string;
-    width?: string;
-    maxHeight?: string;
-    maxWidth?: string;
-    minHeight?: string;
-    minWidth?: string;
     textLength?: number;
+    textChar?: string;
+    blockRatio?: number;
     colorStart?: string;
     colorEnd?: string;
+    variant?: string;
 }
 
 interface StyledSkeletonProps extends SkeletonProps {
     textOnly?: boolean;
+
+    // making nonnullable
+    colorStart: string;
+    colorEnd: string;
 }
+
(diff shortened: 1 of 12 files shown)


src/Skeleton/index.ts
```diff
@@ -1 +1 @@
-export { Skeleton } from './Skeleton.component';
+export { Skeleton, SKELETON_KEY, SKELETON_THEME } from './Skeleton.component';
```
zero0Halo, line 1: I think you were just looking for an excuse to have a variable named `SKELETON_KEY`.

src/Skeleton/Skeleton.component.tsx
```diff
@@ -1,128 +1,130 @@
...
+    colorEnd: string;
 }
 
+export const SKELETON_KEY = 'skeleton';
+export const SKELETON_THEME = {
+    variants: {
+        base: {
+            colorStart: '#E7E7E7',
+            colorEnd: '#D3D3D3',
+        },
```
zero0Halo, line 49: It feels a little weird  hardcoding colors in the component rather than grabbing from the theme, but I get that this is the pattern established by `Button`. Something to think about maybe.

src/Skeleton/Skeleton.component.tsx
```diff
@@ -1,128 +1,130 @@
...
+import {
+    space as spaceStyles,
+    SpaceProps,
+    layout as layoutStyles,
+    LayoutProps,
+    borderRadius as borderRadiusStyles,
+    BorderRadiusProps,
+    variant as createVariant,
+    th,
+} from '@xstyled/system';
```
zero0Halo, line 18: Are the additions of `layout` and `borderRadius` to accommodate rx?

src/Skeleton/Skeleton.component.tsx
```diff
@@ -1,128 +1,130 @@
...
+    const theme = React.useContext(ThemeContext);
+    const { colorStart: themeStart, colorEnd: themeEnd } = colorVariant({
+        ...props,
+        theme,
+    });
 
-    // This is the unicode block character that put together looks like one
-    // long bar. There are shorter and taller characters that could be used
-    // or a fatter or thinner bar. Therefore the bar thickness chould be made
-    // a prop in the future, mapping to different block characters.
```
zero0Halo, line 124: I wouldn't lose all of this comment. I believe `blockRatio` is the response to the second part of the comment, but the first part, `This is the unicode block character that put together looks like one long bar.`, is still helpful. Maybe move this up to where `textChar` is being assigned the character.

src/Skeleton/Skeleton.component.tsx
```diff
@@ -134,17 +136,16 @@ export const Skeleton = ({
 
     const length = textLength || (childText && childText.length);
     const blockText =
-        length &&
-        Array(Math.ceil(length / blockToAverageCharQuotient) + 1).join(
-            blockChar
-        );
+        length && Array(Math.ceil(length * blockRatio) + 1).join(textChar);
+    const textOnly = !!blockText;
```
zero0Halo, line 144: I think this might be over engineering so it take it with a grain of salt (or just a point of discussion); there are a few calculations that always happen when render is triggered. Maybe worth memoizing them? I don't want to flag every spot, but there's a `replace()` with a regex, some array `join()`'s with `Math`, things of that nature. It may not make sense in some cases but I know we're all thinking more about performance so I'm throwing it out there.

zero0Halo (changes requested):
Please do a follow up ticket to update the doc site. The new props and new functionality should be recorded.

src/Skeleton/Skeleton.component.tsx
brandonlilly: Yep, figure its better to use the system

src/Skeleton/Skeleton.component.tsx
brandonlilly: Originally it just couldn't support a theme reference and needed a raw value. This PR actually makes it so it could take `primary.light` for instance. In this case, however, its still raw because I don't think there are appropriate theme colors values to draw from (colors.background.x? doesn't really work). Better to just hardcode imo

src/Skeleton/Skeleton.component.tsx
brandonlilly: I was going to do this but then I decided I want to early return before anything else (so that the normal case of just passing through is as inexpensive as possible). This precludes using a hook like `useMemo`

src/Skeleton/index.ts
brandonlilly: It wasn't intentional but I really enjoy the fact that I had to make one

zero0Halo (approved):
GTG

PR #202, opened 2020-04-07
feat(Button): add spacing props and remove sibling margin

brandonlilly:

---------

Adds spacing props to buttons and removes automatic sibling margins

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@retailmenot/anchor",
-  "version": "1.5.8",
+  "version": "1.6.0",
   "description": "A React UI Library by RetailMeNot",
   "main": "commonjs/index.js",
   "module": "esm/index.js",
--- a/src/Button/Button.component.tsx
+++ b/src/Button/Button.component.tsx
@@ -7,7 +7,12 @@ import styled, {
     FlattenSimpleInterpolation,
     ThemeContext,
 } from '@xstyled/styled-components';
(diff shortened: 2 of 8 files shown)


src/Button/Button.component.tsx
```diff
@@ -370,15 +375,6 @@ const StyledButton = styled('button')<StyledButtonProps>`
     align-items: center;
     outline: none !important;
 
-    margin: ${({ margin }: StyledButtonProps) => margin || `0`};
-    ${({ block, margin }: StyledButtonProps) =>
-        !margin &&
-        css`
-            & + .anchor-button {
```
brandonlilly, line 377: This was designed to automatically space buttons that are side by side. In practice I think it caused more surprise/frustration than it solved, and with space props added we can handle these cases easily/explicitly.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@retailmenot/anchor",
-  "version": "1.5.8",
+  "version": "1.6.0",
```
zero0Halo, line 3: Checking NPM, we actually only have released `1.5.7` so I think leaving it at `1.5.8` is probably fine.

brandonlilly: I bumped up to 1.6.0 because I consider it a "breaking" change. (i.e. Your buttons won't space properly if you were relying on that style)

zero0Halo (approved):
GTG

PR #209, opened 2020-07-07
fix(Dropdown): throttle touchstart and click events

brandonlilly:

---------

Fixes touchstart and click events causing double click of dropdown for mobile devices.

### DIFF
--- a/src/DropDown/DropDown.component.tsx
+++ b/src/DropDown/DropDown.component.tsx
@@ -5,7 +5,13 @@ import classNames from 'classnames';
 import styled, { css } from '@xstyled/styled-components';
 import { space as spaceStyles, SpaceProps } from '@xstyled/system';
 import { fromEvent, Subscription, merge } from 'rxjs';
-import { filter, map, mapTo, distinctUntilChanged } from 'rxjs/operators';
+import {
+    filter,
+    map,
+    mapTo,
+    distinctUntilChanged,
+    throttleTime,
+} from 'rxjs/operators';
 // ANCHOR
 import { get } from '../utils/get/get';
 import { Position } from '../utils/position/position';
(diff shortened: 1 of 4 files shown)


src/DropDown/DropDown.component.tsx
```diff
@@ -140,12 +146,13 @@ export class DropDown extends React.Component<DropDownProps> {
...
             filter(() => this.state.clicked),
-            filter(({ target }) => !dropDown.contains(target))
+            filter(({ target }) => !dropDown.contains(target)),
+            throttleTime(500)
         );
         const dropDownClick$ = merge(
             fromEvent(dropDown, 'click'),
             fromEvent(dropDown, 'touchstart')
-        );
+        ).pipe(throttleTime(500));
```
brandonlilly, line 155: Mobile devices fire both click and touchstart events, so the dropdown was being "clicked" and therefore toggled twice. I'm just throttling it so that it only fires 1 event within 500ms

zero0Halo: Worth making the throttle time a prop so later tweaking (if necessary) won't require another release?

src/Table/Table.component.tsx
```diff
@@ -59,13 +59,14 @@ interface TableProps
...
         BorderProps,
-        WidthProps,
         MaxWidthProps,
         MinWidthProps,
         SpaceProps {
     borderRadius?: string;
     background?: string;
     altBackground?: string;
+    width?: any;
+    height?: any;
```
zero0Halo, line 69: Are these both really `any`? It seems like `string` based on the stories.

src/Table/Table.component.tsx
```diff
@@ -204,12 +205,12 @@ interface TableCellProps
...
         FontWeightProps,
         TextAlignProps,
-        WidthProps,
         MaxWidthProps,
         MinWidthProps,
-        HeightProps,
         SpaceProps {
     align?: 'center' | 'left' | 'right';
+    height?: any;
+    width?: any;
```
zero0Halo, line 213: Same as above.

src/Table/Table.component.tsx
brandonlilly: It's a responsive prop, so it could be a `string` or an `object` (or even a `number`). There was a TypeError here (on master, unrelated to PR) that was pretty painful, so I chose not to fight it.

src/DropDown/DropDown.component.tsx
brandonlilly: Almost everything I read online said there was ~300-350ms delay between the events. I went for 500 to be safe.

Leaning against prop because if I'm wrong about my assumptions, I'll need to fix the defaults here anyway. Prop also a bit weird because the throttle times are set in the componentDidMount, so it wouldn't respond to prop value changes over the life of the component.

zero0Halo: I'd be less concerned about the prop changing as much as having a way to allow configuration of the throttle. Not a hill to die on though.

zero0Halo (approved):

PR #217, opened 2020-08-27
fix(Pagination): controlled initial value and onChange fire logic

brandonlilly:

---------

In a controlled Pagination, initial values that weren't `1` would be overridden. onChange and onTotalChange would extraneously fire on initial render. This PR fixes both of those problems.

### DIFF
--- a/src/utils/useUpdateEffect/useUpdateEffect.ts
+++ b/src/utils/useUpdateEffect/useUpdateEffect.ts
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-arrow-callback */
 import * as React from 'react';
 
 // useUpdateEffect is a useEffect that doesn't fire on initial render
--- a/src/utils/useUpdateEffect/useUpdateEffect.ts
+++ b/src/utils/useUpdateEffect/useUpdateEffect.ts
@@ -1,0 +1,43 @@
+import * as React from 'react';
+
+// useUpdateEffect is a useEffect that doesn't fire on initial render
+// for detecting when values change
+// https://stackoverflow.com/a/58217148
+
+const useIsMounted = function useIsMounted() {
+    const isMounted = React.useRef(false);
+
+    React.useEffect(function setIsMounted() {
+        isMounted.current = true;
+
+        return function cleanupSetIsMounted() {
+            isMounted.current = false;
+        };
+    }, []);
+
+    return isMounted;
+};
+
+export const useUpdateEffect = function useUpdateEffect(
+    effect: () => void | (() => void),
+    dependencies: Array<any>
+): void {
+    const isMounted = useIsMounted();
+    const isInitialMount = React.useRef(true);
+
+    React.useEffect(() => {
+        let effectCleanupFunc = function noop() {};
+
+        if (isInitialMount.current) {
+            isInitialMount.current = false;
+        } else {
+            effectCleanupFunc = effect() || effectCleanupFunc;
+        }
+        return () => {
+            effectCleanupFunc();
+            if (!isMounted.current) {
+                isInitialMount.current = true;
+            }
+        };
+    }, dependencies); // eslint-disable-line react-hooks/exhaustive-deps
+};
--- a/src/utils/useUpdateEffect/useUpdateEffect.ts
+++ b/src/utils/useUpdateEffect/useUpdateEffect.ts
@@ -4,6 +4,7 @@ import * as React from 'react';
 // for detecting when values change
 // https://stackoverflow.com/a/58217148
 
+// Named functions are for stack traceability
 const useIsMounted = function useIsMounted() {
     const isMounted = React.useRef(false);
 
@@ -20,12 +21,13 @@ const useIsMounted = function useIsMounted() {
 
 export const useUpdateEffect = function useUpdateEffect(
     effect: () => void | (() => void),
-    dependencies: Array<any>
+    dependencies: any[]
 ): void {
     const isMounted = useIsMounted();
     const isInitialMount = React.useRef(true);
 
     React.useEffect(() => {
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
         let effectCleanupFunc = function noop() {};
 
         if (isInitialMount.current) {
@@ -39,5 +41,5 @@ export const useUpdateEffect = function useUpdateEffect(
                 isInitialMount.current = true;
             }
         };
-    }, dependencies); // eslint-disable-line react-hooks/exhaustive-deps
+    }, dependencies);
 };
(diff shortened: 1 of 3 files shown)


src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+
+// useUpdateEffect is a useEffect that doesn't fire on initial render
+// for detecting when values change
+// https://stackoverflow.com/a/58217148
+
+const useIsMounted = function useIsMounted() {
```
zero0Halo, line 9: I'm struggling to understand what's going on here, the assignment of a named function to the same-named variable. I'd think that the function declaration name would either be different from the variable it's assigned to so that it could use its own name for recursion, or... well I'm not sure what else for. :)

Is it for hoisting for some reason?

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+    const isMounted = React.useRef(false);
+
+    React.useEffect(function setIsMounted() {
```
zero0Halo, line 12: eslint is not liking the named function instead of an anonymous one. I normally would name a function if it was resolving an event and I wanted to remove the event listener, but what's this doing?

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+        isMounted.current = true;
+
+        return function cleanupSetIsMounted() {
```
zero0Halo, line 15: Again, eslint is not happy about the named function.

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+            isMounted.current = false;
+        };
+    }, []);
+
+    return isMounted;
+};
+
+export const useUpdateEffect = function useUpdateEffect(
+    effect: () => void | (() => void),
+    dependencies: Array<any>
```
zero0Halo: eslint: `Array type using 'Array<any>' is forbidden. Use 'any[]' instead`

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+        } else {
+            effectCleanupFunc = effect() || effectCleanupFunc;
+        }
+        return () => {
+            effectCleanupFunc();
+            if (!isMounted.current) {
+                isInitialMount.current = true;
+            }
+        };
+    }, dependencies); // eslint-disable-line react-hooks/exhaustive-deps
```
zero0Halo: eslint: `Definition for rule 'react-hooks/exhaustive-deps' was not found.eslint(react-hooks/exhaustive-deps)`

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const useUpdateEffect = function useUpdateEffect(
+    effect: () => void | (() => void),
+    dependencies: Array<any>
+): void {
+    const isMounted = useIsMounted();
+    const isInitialMount = React.useRef(true);
+
+    React.useEffect(() => {
+        let effectCleanupFunc = function noop() {};
```
zero0Halo, line 32: eslint is not pleased about the empty `noop` function.

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        return function cleanupSetIsMounted() {
+            isMounted.current = false;
+        };
+    }, []);
+
+    return isMounted;
+};
+
+export const useUpdateEffect = function useUpdateEffect(
```
zero0Halo, line 23: Again, unclear on the same-name function assignment.

src/utils/useUpdateEffect/useUpdateEffect.ts
brandonlilly: My understanding is that this is a convention some people use because it is convenient for stack tracing. You name your function with the same name so it isn't merely anonymous.

src/utils/useUpdateEffect/useUpdateEffect.ts
brandonlilly: Lol, I tried `Any[]` but couldn't remember the right syntax. Ty

src/utils/useUpdateEffect/useUpdateEffect.ts
brandonlilly: annoying that the local `yarn lint` apparently isn't the linter the runners are using

src/utils/useUpdateEffect/useUpdateEffect.ts
zero0Halo: Ok, I can see that. Could you add a comment to that effect? I'm probably not the only person who will go, "Whaaaat?"

src/utils/useUpdateEffect/useUpdateEffect.ts
zero0Halo: Based on your above comment, I'm guessing stack trace?

src/utils/useUpdateEffect/useUpdateEffect.ts
zero0Halo: I get it now. I think it's not necessary to put a comment for every instance this syntax is used as long as there's one near the top explaining it.

src/utils/useUpdateEffect/useUpdateEffect.ts
zero0Halo: Very. Going to have to look into that.

src/utils/useUpdateEffect/useUpdateEffect.ts
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
```
zero0Halo, line 2: If you add `/* eslint-disable prefer-arrow-callback */` here (line 1) that'll kill the eslint errors about preferring arrow functions

brandonlilly: I'm very confuse. Where are you seeing these eslint errors?

brandonlilly: Afaict all the CI has passed

zero0Halo: Looks like this is an issue with VSC on my machine, so going to resolve this and move along.

zero0Halo (approved):
Game on.

